PR #25, opened 2017-01-22
Make libsweep compilable with a C++ compiler

PeterJohnson:
Currently the sources use GNU extensions and C99-only features, so are not compilable with a standard C++ compiler.  This PR fixes this.

Also, this PR renames time.h to sweep_time.h to avoid issues with include paths and name conflicts with the standard library time.h header.


libsweep/protocol.c
```diff
@@ -55,14 +63,15 @@ static uint8_t sweep_protocol_checksum_response_scan_packet(sweep_protocol_respo
...
   SWEEP_ASSERT(cmd);
   SWEEP_ASSERT(error);
 
-  sweep_protocol_cmd_packet_s packet = {.cmdByte1 = cmd[0],    //
-                                        .cmdByte2 = cmd[1],    //
-                                        .cmdParamTerm = '\n'}; //
+  sweep_protocol_cmd_packet_s packet;
+  packet.cmdByte1 = cmd[0];    //
+  packet.cmdByte2 = cmd[1];    //
+  packet.cmdParamTerm = '\n';  //
```
daniel-j-h: Here and below: feel free to remove the `//` - they were used because `clang-format` can be a bit ugly from time to time.

PeterJohnson:
The main reason was to make it easier to vendorize for FRC use, as a C++ compiler is always used in the vendor template (https://github.com/PeterJohnson/sweep-sdk).  The vendorized build makes it very easy to integrate into eclipse as teams can just unzip it rather than needing to build it themselves, plus it provides Java wrappers.

There also doesn't seem to be a lot of benefit to doing the C99 things like piecewise initialization in this case.

daniel-j-h:
Sweet, love it!

If you maintain the integration feel free to drop a link into this repository's main readme, e.g. under a "who's using this" section.

Also do you have any plans to upstream the Java wrapper? How big of a lift would that be? The main reason we don't have one is my lack of experience and use-cases in that regard.

PeterJohnson:
Ok, I definitely plan to keep it up to date with changes here.  Upstreaming a variant of the Java wrapper is possible, although the version in my repo is FRC specific so it wouldn't be identical.

Issue #1, opened 2016-11-01
Support for macOS


daniel-j-h:
Currently I only tested `libsweep`, `SweepPy` and `SweepJs` on my Linux Ubuntu machine.

For platform specifics we can dispatch in [config.mk](https://github.com/scanse/sweep-sdk/blob/f7fbfbd2e304887a23d13678b40e64ecff770485/libsweep/config.mk). I assume we have to change some specifics such as the installation paths, maybe some cflags/ldflags. The `_unix` implementations are POSIX-conforming and should Just Work (tm).

daniel-j-h:
Here's a macOS build running on Windows: https://github.com/scanse/sweep-sdk/pull/29

Surprisingly enough it Just Works (tm). At least compiling and linking the dummy implementation.

The Sweep comes with a FTDI USB serial device converter which a friend of mine said should work just fine on macOS passing in a dev path such as `/dev/cu.usbserial0233`.

I can't test this and would love for someone on macOS to try it out with a real device.

dcyoung:
I gave the travis-osx build a shot on OSX El Capitan. 

Obviously the hardcoded serial port name [here](https://github.com/scanse/sweep-sdk/blob/033d4a6166701cef4adb80161055d407a9a1cbf1/libsweep/src/sweep.cc#L52) had to be changed out in favor of the appropriate /dev/cu.* for the device. The only other issue I encountered was the baud rate. The get_baud function was falling through to the default error case [here](https://github.com/scanse/sweep-sdk/blob/033d4a6166701cef4adb80161055d407a9a1cbf1/libsweep/src/unix/serial.cc#L239), and B115200 wasn't identifiable when i tried it manually (ie: `speed_t baud = B115200;`). 

I just went ahead and hard coded the value by changing this line [here](https://github.com/scanse/sweep-sdk/blob/033d4a6166701cef4adb80161055d407a9a1cbf1/libsweep/src/unix/serial.cc#L320) to `speed_t baud = 115200;` and everything worked just fine. Library built and installed and the c and c++ examples ran as expected. I'll try to test the bindings now.

daniel-j-h:
Sweet!

The dev path auto-detection was something we wanted to remove anyway.

The Node binding build system lacked support for macOS - I just added it in https://github.com/scanse/sweep-sdk/pull/29.

daniel-j-h:
[...]

PR #29, opened 2017-01-24
Adds Travis macOS Build

daniel-j-h:
Will probably fail - let's see where and why.

Also let's keep the PR open until we have macOS builds working. cc @dcyoung

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: generic
 sudo: required
 dist: trusty
 
-
 matrix:
   fast_finish: true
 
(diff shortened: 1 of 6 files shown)


daniel-j-h:
node seems to need manual installation on macOS, something along the lines of

```
git clone --depth 1 --branch v0.33.0 https://github.com/creationix/nvm.git
. nvm/nvm.sh
nvm install 4
nvm use 4
```

should do the trick.

We should probably test LTS v4 and LTS v6 (https://github.com/nodejs/LTS#lts_schedule) on all builds.

daniel-j-h:
Running out some minor fixes here, e.g. the `libsweep.dylib` (macOS) vs `libsweep.so` (Linux) difference in the Python bindings which we also would have hit in the Windows port.

Otherwise it seems like at least the dummy implementation and all the examples compile and link without problems. Seems like the blocker here is the serial implementation for macOS.

daniel-j-h:
Merging fixes into master. After all the dummy impl. is all we are ever able to test on Travis.

@dcyoung found some issues with the real impl. on macOS - see discussion in https://github.com/scanse/sweep-sdk/issues/1.

PR #32, opened 2017-01-25
Stops scanning and flushes serial device on destruction, resolves #23

### LINKED ISSUE #23: Examples Error on alternating runs without device reset
Running the examples `./example-c`, `./example-c++` or `./example-viewer` works the first time, but unless the sensor is reset between each run, the example will error on subsequent attempts in an alternating pattern. Ie: the 2nd attempt will fail, but the 3rd will succeed, 4th will fail, etc

In the case of `./example-c` the stated error is:

> Error: unable to receive motor speed command response

In the case of `./example-c++` the stated error is:

> Error: unable to receive stop scanning command response

And in the case of `./example-viewer` the stated error is:

> Error: unable to receive start scanning command response

Has this always been the behavior or is it a new issue? I don't believe any changes were made to the protocol for `DS` or `DX` receipts. I haven't started tracking down the origin of the issue yet, but I'm assuming the device is not being shutdown properly. Whether its a logical error in the example itself, or an error with `libsweep` is unclear.

Edit: Looking at the example code, I don't see where we send a `DX` stop command to the sensor. However, after the example runs the device does stop transmitting serial data (judging from the TX led on the USB Serial adapter which stops blinking). It was my understanding that the device would continue to transmit data regardless of the state of the host port.... so somehow the device is getting that stop command. Is it hidden somewhere I might have overlooked?  

If the example does trigger the host to send a `DX` stop data command before or during the process of shutting down, are the still incoming `Data Block` receipts handled properly? Recall that after sending a `DX` command, the host should still expect to receive `Data Block` receipts until the `DX` receipt is received, which indicates no further `Data Block` receipts are coming. Is it possible that the host sends the `DX` and then immediately shuts down? This might be leaving incoming bytes in the serial buffer that are misinterpreted at the beginning of the example's next execution... producing the observed alternating errors.

daniel-j-h:
Calling the stop function on the user-side is not really necessary as we're doing it on device destruction anyway if it wasn't done so far. It's more to get users familiar with the symmetric API.

I confirmed #23 being an issue with the recent units and this changeset fixing it for me.

@dcyoung can you confirm? Edit: confirmed working in issue.

### DIFF
--- a/libsweep/examples/example.c
+++ b/libsweep/examples/example.c
@@ -62,6 +62,10 @@ int main() {
     sweep_scan_destruct(scan);
   }
 
+  // Stop capturing scans
+  sweep_device_stop_scanning(sweep, &error);
+  check(error);
+
   // Shut down and cleanup Sweep device
   sweep_device_destruct(sweep);
   return EXIT_SUCCESS;
--- a/libsweep/examples/example.cc
+++ b/libsweep/examples/example.cc
@@ -17,6 +17,9 @@ int main() try {
       std::cout << "angle " << sample.angle << " distance " << sample.distance << " strength " << sample.signal_strength << "\n";
     }
   }
+
+  device.stop_scanning();
+
 } catch (const sweep::device_error& e) {
   std::cerr << "Error: " << e.what() << std::endl;
 }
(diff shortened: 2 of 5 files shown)


dcyoung:
@daniel-j-h  We're getting a bit cluttered with branches. Do you have any issue with deleting this `stop-scanning-flush` branch now that it has been merged?

Issue #31, opened 2017-01-25
Worker Thread Discussion


dcyoung:
We've discussed the need to introduce a worker thread for IO. However, the implementation must consider cross-platform compatibility. The debate is whether to...

1. Shift the `libsweep` library's implementation to C++, while still maintaining the C interface. The C++ implementation would allow for all threading to handled by C++11 standard thread library, and avoid the need for thread  abstraction. This would both consolidate the code. @daniel-j-h  You mentioned that this requires us to link in libstdc++/libc++ which is a pain for abi stability and shipping portable binaries. Could you elaborate on those two points a bit? Isn't shipping platform specific binaries pretty standard? 

2. Abstract the threading into a shared interface that is platform agnostic, and then provide platform specific implementations. 

For option 2 (abstraction), we could do this in a similar fashion to `serial.h` + (`serial_unix.c` or `serial_win.c`). In this case, we'd have something like `sweep_threading.h` + (`sweep_threading_unix.c` or `sweep_threading_win.c`). This interface would define an abstracted thread, mutex, conditional var, and single producer single consumer queue. Then the methods from the `serial` and `sweep` files would make use of the threads and queues via this abstracted interface. The unix implementation could use pthreads. I'm still open to suggestions for the win implementation but it could use WINAPI. Using WINAPI would require the inclusion of <windows.h>. Any complications with the build process or bindings there? @daniel-j-h 

What do we think about the pros/cons of both?

Given that the current style of sweepSDK is to avoid global state...

> No global state: _s context objects (_s since POSIX reserves _t) hold state and have to be passed explicitly.

how should we store and reference the working queue? I spent a while today trying to implement a queue and a worker thread that conformed to the limitations of no global state and the Pimpl idiom, but I didn't get very far.... I'm still getting the hang of the Pimpl idom :/ @daniel-j-h  if you've got a plan for how to accomplish this I'm happy to collaborate however I can.

daniel-j-h:
The pain point with libstdc++ / libc++ is as follows:

if we ever want to ship pre-built binaries (for example to let users do `npm install sweepjs` and not have a compiler around for it to work) practically speaking what we have to do is dig out a old Linux distribution and link against its C++ stdlib.

libstdc++ is forward compatible using versioned symbols: we can build against older versions and still safely run against a newer version. This also means we have to find a libstdc++ version which already is C++11 feature complete (for threads) and at the same time old enough for us to support newer versions.

[...]

PR #33, opened 2017-01-27
Uses C++11 for Implementation

daniel-j-h:
Experimental: for discussion in https://github.com/scanse/sweep-sdk/issues/31.

Impl. is C++11 now. Needs to be fully converted to C++ idioms, e.g. using `new` / `delete`, `auto` etc.

cc @dcyoung


daniel-j-h:
Just rebased against master and had a look here. We can merge this and adapt to C++11 idioms later or already do it in one go. This is what comes to mind from a first glance:
- rip out time header / impl. - use [sleep_for](http://en.cppreference.com/w/cpp/thread/sleep_for) from the stdlib (need to link in pthread and include thread)
- [ ] use `auto` instead of explicitly specifying types as in `auto p = new T;`
- [ ] use `new` and `delete` or `new []` and `delete []` for opaque pointers in public API
- [ ] do not use manual memory management in impl. - use `std::vector` and RAII friends
- [ ] let impl. throw exceptions, in `sweep.cc` catch all of them and transform to errors
- [ ] use `static_assert` for protocol type sizes (check Windows)

MikeGitb:
If I may make a suggestion: Currently the c++ class `sweep` seems to be a wrapper around the c-interface. Now, if you want to implement the c-interface via c++, wouldn't it make sense to rewrite the `sweep` class more or less from scratch and make the c-interface a warper around that class (Effectively making `sweep_device_s` a typedef for `sweep*` )? 

On the other hand: Except for better cross-platform support (threads and chrono in particular): Are there any real benefits in switching the implementation to c++? I'm certainly not against it and will only use the library in c++ projects anyway, but there might be a price to pay in terms of binary compatibility and supported platforms (I don't have any experience with that aspect though) .

libsweep/config.mk
```diff
@@ -5,9 +5,9 @@ UNAME := $(shell uname)
 
 ifeq ($(UNAME), Linux)
   PREFIX ?= /usr
-  CFLAGS += -O2 -Wall -Wextra -pedantic -std=c99 -Wnonnull -fvisibility=hidden -fPIC -pthread
+  CXXFLAGS += -O2 -Wall -Wextra -pedantic -std=c++11 -fvisibility=hidden -fPIC -fno-rtti -fno-exceptions -pthread
   LDFLAGS += -shared -Wl,-soname,libsweep.so.$(VERSION_MAJOR)
-  LDLIBS += -lpthread
+  LDLIBS += -lstdc++ -lpthread
```
MikeGitb: Are you sure you have to explicitly mention the standard library here? I'd leave choosing the standard library and compiler to the user

daniel-j-h:
See discussion in https://github.com/scanse/sweep-sdk/issues/31. The main points are us not having to abstract timing and threading features. Also down the line we want to provide a utility library for things like filters, clustering, etc. which will most likely require C++ anyway.

The reason we only switch the implementation to C++ but keep the interface behind `extern "C"` blocks is to allow for FFI bindings without the hassle to go through Swig / Boost.Python and related. This also allows us to provide ABI stability. On top of this stable C library we provide a convenience C++ header for users.

Check out http://www.slideshare.net/StefanusDuToit/cpp-con-2014-hourglass-interfaces-for-c-apis (there should also be a corresponding video from that talk)

libsweep/config.mk
daniel-j-h: Yes we have to link in the stdlib - at least as soon as we make use of C++ features such as `std::thread` in the implementation.

We could make the compiler and stdlib more easily configurable. On the other hand libstdc++ & gcc is the default on Linux and Clang & libc++ the default on macOS. If users want to change this the `config.mk` file is easily editable.

MikeGitb:
I understand the reason for providing a C interface. What I don't understand is why the c++ convenience headers should use that interface instead of the (future) c++ implementation directly.

daniel-j-h:
Because the C++ implementation does not provide any ABI stability guarantees internally. It is the C API on top of the C++ ABI providing these guarantees.

Here's a use-case: you don't have a device yet, so you `make dummy` and get a `libsweep.so` which always returns static point cloud data. Now you can already build your application / bindings and include the C API header and link to `libsweep.so`.

Once you're device arrives all you have to do is `make` in order to get the real `libsweep.so` working with your device. The ABI is the same - only the internal implementation changes - which means you can just swap out the shared library and everything keeps working.

Similar use-case is when updating from one minor release to the next, bug fixes etc.

daniel-j-h:
To provide an additional example: the `example-viewer.cc` realtime viewer uses the C++ convenience header and SFML2 for visualization. You can run it with the dummy and real `libsweep.so` and swap them out behind the back. The `example-viewer` binary does not care since the ABI is the same.

daniel-j-h:
@dcyoung I just pushed some changes: transitioning to auto, new/delete, namespaces and statically asserted in-memory packet sizes. Care to have a look? I want to merge this in soon, since all other pull requests basically depend on this.

What can be done later are features such as using exceptions in the implementation and catching them at the C API boundary. I'm thinking of

```c++
// Base error the impl. throws
struct error : std::runtime_error {
  using base = std::runtime_error;
  using base::base;
};

struct serial_error : error { using error::error };
struct protocol_error : error { using error::error };
```

then the subsystems no longer take `error*`, fill it, and let the call site check this pointer.
Instead, the subsystems throw specific exceptions. Code would change from

```c++
serial::error_s error = nullptr;

serial::device_read(serial, args, &error);

if (error) {
  handle(error);
  serial::error_destruct(error);
}
```

To

```cpp
serial_device_read(serial, args);
```

with all functions in `sweep.cc` wrapped with

```c++
void fn() try {
  // impl.
} catch (const sweep::error& e) {
  // don't let exception propagate API boundary
}
```

At the API boundary (in `sweep.cc`) we catch-all `catch (const sweep::error&)` and translate it to errors for the user. This would require us compiling with exceptions for the impl. (see `-fno-exceptions` in cxx flags).

MikeGitb:
Are you sure that swapping out the library without recompiling the c++ application will be a common requirement? Personally I find it much more likely that the library ends up as part of the source tree of the final application (especially in case of embedded applications). And if one wants to profit from changes in the convenience headers (which are part if the library?) he has to recompile it anyway. 
My main concern main with the current design is efficiency and code complexity: E.g.: Even if the internal and external representation of a scan are identical, every single value has to be copied via an opaque function call and you are ending up allocating the memory at least twice. The same goes for errors (although efficiency isn't necessarily a problem here).

Regarding the use of exceptions : Don't you have to catch all exceptions at the API-boundaries? Not just the ones derived from `sweep::error`?

daniel-j-h:
The main design goal is having a stable ABI all other bindings can work on top of it.

If you are concerned with efficiency or memory allocation, feel free to open an issue re. support for allocators. Here's a good overview http://lucumr.pocoo.org/2013/8/18/beautiful-native-libraries/

Then users could plug in e.g. their stack allocators if needs be. (I'm not sure what's your issue here, though. One allocation per full 360 scan seems totally fine for every environment we want to target with this library)

Re. exceptions, yes you are right. Although the only exception which should get thrown otherwise should be `std::bad_alloc` and terminating on that is a reasonable choice.

daniel-j-h:
Merging this to unblock other work - we can work on the remaining smaller issues independently.

PR #34, opened 2017-01-27
Changed signal strength storage format to match all other Scanse docs…

dcyoung:
… and software. 

### Signal Strength Format
The existing implementation stored signal strength as an int in the range 0:100. While this pseudo-normalized format is a conceptual improvement for those unfamiliar with a byte, it is not congruent with the other software and documentation provided by Scanse. To avoid confusion, we should keep it the same. If we wanted to go the normalized route, we might consider normalized float between 0:1. But given the SDK design goals (passing only signed ints) this is better left to the user.

### Clarify storage technique
Additionally, the method by which sensor data is converted and stored in signed ints is somewhat cryptic for the complete novice Sweep user. Given that the SDK may serve as an example of parsing data, this PR adds unnecessary but explicit casts & comments to improve understanding when visually scanning the code. 

### Examples print familiar values
This PR also modifies the examples to print familiar values (ie: degrees, instead of milli-degrees)

### DIFF
--- a/libsweep/examples/README.md
+++ b/libsweep/examples/README.md
@@ -14,7 +14,7 @@ To build:
     mkdir build
     cd build
     cmake ..
-    make
+    cmake --build .
 ```
 
 **Note:** the viewer requires SFML2 to be installed.
--- a/libsweep/examples/README.md
+++ b/libsweep/examples/README.md
@@ -14,7 +14,7 @@ To build:
     mkdir build
     cd build
     cmake ..
-    cmake --build .
+    make
 ```
 
 **Note:** the viewer requires SFML2 to be installed.
--- a/libsweep/examples/example.cc
+++ b/libsweep/examples/example.cc
@@ -14,7 +14,8 @@ int main() try {
     const sweep::scan scan = device.get_scan();
 
     for (const sweep::sample& sample : scan.samples) {
-      std::cout << "angle " << sample.angle << " distance " << sample.distance << " strength " << sample.signal_strength << "\n";
+      std::cout << "Angle: " << sample.angle / 1000.f << ", Distance: " << sample.distance << ", Strength "
+                << sample.signal_strength << "\n";
     }
   }
 
--- a/libsweep/examples/example.cc
+++ b/libsweep/examples/example.cc
@@ -14,7 +14,7 @@ int main() try {
     auto scan = device.get_scan();
 
     for (auto sample : scan.samples) {
-      std::cout << "angle " << sample.angle << " distance " << sample.distance << " strength " << sample.signal_strength << "\n";
+      std::cout << "Angle: " << sample.angle/1000.f << ", Distance: " << sample.distance << ", Signal Strength: " << sample.signal_strength << "\n";
     }
   }
 
--- a/libsweep/sweep.c
+++ b/libsweep/sweep.c
@@ -23,18 +23,18 @@ typedef struct sweep_device {
  * Struct: sweep_scan
  * ------------------
  * Contains the data for sensor readings in a scan
- * 
+ *
  *  angle: array of azimuth values (1 for each reading)
  *        (in millidegrees, ie: 241.375deg stored as 241375millideg)
  *  distance: array of range values (1 for each reading)
  *        (in centimeters)
  *  signal_strength: array of signal strength values (1 for each reading)
  *        (integer values between 0:255, higher is better)
- *  count: number of sensor readings in the scan 
+ *  count: number of sensor readings in the scan
  */
 typedef struct sweep_scan {
-  int32_t angle[SWEEP_MAX_SAMPLES]; // in millidegrees 
-  int32_t distance[SWEEP_MAX_SAMPLES]; // in cm
+  int32_t angle[SWEEP_MAX_SAMPLES];           // in millidegrees
+  int32_t distance[SWEEP_MAX_SAMPLES];        // in cm
   int32_t signal_strength[SWEEP_MAX_SAMPLES]; // range 0:255
   int32_t count;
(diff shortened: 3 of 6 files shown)


libsweep/examples/README.md
```diff
@@ -14,7 +14,7 @@ To build:
     mkdir build
     cd build
     cmake ..
-    cmake --build .
+    make
```
daniel-j-h: Can you revert this change - cmake abstracts over the build system in use. On Linux the default generator is Unix Makefiles, so make is fine there. This is not the case everywhere. Also no the case when changing the generator, e.g. with `-G Ninja`.

libsweep/examples/example.cc
```diff
@@ -14,7 +14,7 @@ int main() try {
     auto scan = device.get_scan();
 
     for (auto sample : scan.samples) {
-      std::cout << "angle " << sample.angle << " distance " << sample.distance << " strength " << sample.signal_strength << "\n";
+      std::cout << "Angle: " << sample.angle/1000.f << ", Distance: " << sample.distance << ", Signal Strength: " << sample.signal_strength << "\n";
```
daniel-j-h: Can you clang-format all your changes - there is a `.clang-format` style file in the repo.

Work with `clang-format-3.8` for sure - other versions might introduce some differences - we have to see.

If this is too much of a hassle for you I can do it before merging.

libsweep/sweep.c
```diff
@@ -19,10 +19,23 @@ typedef struct sweep_device {
...
+ * Contains the data for sensor readings in a scan
+ * 
+ *  angle: array of azimuth values (1 for each reading)
+ *        (in millidegrees, ie: 241.375deg stored as 241375millideg)
+ *  distance: array of range values (1 for each reading)
+ *        (in centimeters)
+ *  signal_strength: array of signal strength values (1 for each reading)
+ *        (integer values between 0:255, higher is better)
+ *  count: number of sensor readings in the scan 
+ */
```
daniel-j-h: Hm care to remove the comment here - the public API docs should be in the README.md. This is the impl. file.

libsweep/examples/README.md
dcyoung: Sounds good. Will do.

dcyoung:
This pull request was more a chance do test the travis system for me. It wasn't very properly formatted. I'll reformat it and submit it again. 

Regarding comments in general.... I mentioned the comments might be useful to a novice user, but they were also a method of documenting the internals for my own reference.

Obviously I agree that we should have a very well documented API in the README. I tend to be verbose in the implementation comments as well, even if just for my own reference. I agree that we could leave function descriptions out of the implementation files, but I'd definitely like to see them in the header files at the very least. 

We could also adopt a consistent comment style for auto-generating API docs. This isn't really an issue with this SDK in isolation, but thinking forward to all the extensions and libraries... I'd prefer to have a consistent style. Would function descriptions in the header files serve a happy medium?

libsweep/examples/example.cc
dcyoung: Look like my IDE supports specifying a clang-format binary to use. I'll try to grab 3.8 and auto format everything.

libsweep/sweep.c
dcyoung: Obviously we want to document the API in the ReadMe... but I'd also definitely like to have docs in the code for my own reference more than anything else. For things like function definitions, we could put comment descriptions in the header files, but for these structs... the public entity is just the opaque pointer. How would you recommend documenting these structs in code, if not here? I'm open to suggestions.

daniel-j-h: Hm we only have a few internal structs and so far I felt no need to document these.

If you want to, I would recommend adding inline comments to the struct members, such as

```c++
    // Number of samples in the scan.
    int32_t num_samples;
    ..
```

daniel-j-h:
I had bad experiences with auto-generated documentation so far. Across several languages (C, C++, Python, Javascript, ..) this becomes close to impossible. The good thing here is our public API should be stable and not change much.

If you want to I can add ReadTheDocs integration, provide a Sphinx config, `index.rst` and let it automatically render our docs. Feel free to open a ticket for this.

My main issue with this PR is still the change in semantic.

(Note: it seems like you merged master into your branch for updating. What you should do instead is `git rebase master` to rebase your work on top of latest master. If you have conflicts, resolve them, `git add`, `git rebase --continue`)

MikeGitb:
Semantical changes aside: What would be a *sensible* resolution, given the hw's capabilities and the data transmitted over the serial interface. If you want to pass integers instead of floats, that is ok (especially if used on microcontrollers) but the main question for me would be: What resolution do I have to choose to not loose any information.

daniel-j-h:
Not sure what you mean here. The signal strength is a [uint8_t](https://github.com/scanse/sweep-sdk/blob/bdcf9e75da0fdd904880fcb3f1c7eda5851be672/libsweep/protocol.h#L70). The normalization to 0--100 was just a first take in abstracting away / hiding the underlying byte range from the user. Shrinking 0--256 into 0--100 is less precise for sure, but then again I'm not sure if we need the precision here.

dcyoung:
Having looked more into the design philosophy behind this SDK, I see your logic and I'll go ahead and remove the comments.

However, I'd definitely like to deliver a value between 0:255 to the user. All other docs, manuals and protocol spec discuss the signal strength value as 0:255. I think introducing a different normalization is unnecessary. There are relatively few people who have received the device yet, so I don't mind breaking the API this early on. Its definitely worth it to keep everything consistent.

daniel-j-h:
Okay, I'm :+1: with changing to 0--256 now. Better to do it sooner than later.

PR #36, opened 2017-01-28
avoid using auto in example.cc

matpalm:
explicit info on types is part of the reason people read examples IMHO


daniel-j-h:
Looks good. I would take the objects by value. By const ref is of no benefit here since the API returns by value anyway.

daniel-j-h:
Merged - thanks for your contribution!

PR #39, opened 2017-02-03
[IGNORE] --old    Windows Support

dcyoung:
### EDIT: 
There were too many conflicting goals (c++, file structure, win, visual studio etc), so I made a new PR #44 built off master, post C++ merge. We can ignore this old PR, but I figured I'd leave it here in case we want to reference the discussion.

### Old... for reference
Adds windows support via MinGW. Should be merged after #38  (file-structure PR).

- Provides windows implementation for sweep_time file.
  - Sleep function from the winAPI is only capable of specifying millisecond accuracy, so the `sweep_sleep_microseconds` method was altered and renamed in order to take milliseconds instead of microseconds. This change affects `sweep_time.h` and `time_unix.c` as well.
- Provides windows implementation of serial file.
- Modifies `config.mk` and `Makefile` to support building on windows (currently only supports MinGW+MSYS)
- Modifies `sweep_device_construct_simple` to take the port name. This is necessary given that port name will vary by system and device. It also permits multiple devices via simple method. I'm not sure if there is a way around specifying the port name. Perhaps we could query the OS and provide a default for each, but even then port names vary by device instance on windows.  **NOTE:** this change breaks the current API.
  - SweepJS bindings were updated. 
  - Python bindings were NOT updated, although the python script does not error on travis. Do they still need modification @daniel-j-h 
  - Modified the examples to provide port name
- Removes the viewer example from the build list temporarily... this needs a solution.


daniel-j-h:
Sweet!

Not sure about the `construct_simple` change. We know the path at compile-time, right? There should be no need to change the API, instead we can dispatch based on OS in the build system and compile in one path or the other. The hacky way would be to ifdef the path out. The proper way would probably be to provide a header with an `extern const char *` and link in the right object file defining the default path (check protocol, where this is done for the constants).

Another option would be to add a construct function taking only a path and keeping the existing ones.

If the default path even varies on Windows, the user should use the regular `construct` path and specify it. Maybe a Windows specific example could help there.

We should merge the C++11 impl. pull request first.
The you can remove the timing feature and switch to chrono.

daniel-j-h:
We should also integrate a Windows specific CI service.
Travis handles Linux and macOS only, we can use AppVeyor for this.

MikeGitb:
> We know the path at compile-time, right? 

maybe a stupid question, but how? Doesn't this depend on which other devices are currently connected?

daniel-j-h:
Sure, but it's the same for Linux.

The idea behind the `_simple` constructor is to make some guesses and hopefully everything will work out fine in most of the cases. That's why I said we probably want to provide the following overloads

```c++
construct(); // guess devpath, bitrate
construct(devpath); // guess bitrate
construct(devpath, bitrate); // user is explicit
```

to keep the ABI as is and to only add an additional constructor.

dcyoung:
Hmmm, I see how this assumption is "safer" on Linux, but on windows the name of the com port is assigned the first time the device is plugged in. So even if it is the only serial device currently connected, it will not default to a specific COM#. And many sweep users will likely have numerous serial devices. I don't think it is a good idea to try and assume a default port name on windows.  

I strongly believe in the design philosophy of providing simple methods that make assumptions about what kind of settings the user wants to use. But port name is not a setting, it is a target. An assumed port name is NOT guaranteed to exist, nor is it guaranteed to be a Sweep device. I'm uncomfortable providing "simple" methods which can't guarantee their assumptions will work. 

While we could just direct windows users to call the non-simple method, this starts delineating the API by OS. I do NOT want to provide methods that only work, or work significantly better, on specific operating systems.  

I don't think users will have an issue specifying a port name. I think this is fairly standard. Additionally, specifying the port name allows the simple method to be used when connecting multiple devices.

daniel-j-h:
I see, okay. Let's do it this way then:
- Provide the three constructors as described above. For Windows and Linux put in a default dev path that is _somewhat_ reasonable.
- Document the new constructor in the README and make clear the user should specify the dev path if possible and that the default is just a guess.
- Explicitly specify paths in the examples.

Thinking about it we never should have provided the auto-guessing `construct()` default constructor with these problems on Windows. Or is there an easy way to iterate over all dev paths and check if it's a Sweep?

dcyoung:
I do not believe there is a way to check that a serial device is a sweep without attempting to send it a command and listening for the proper response. You could theoretically make a better assumption by querying serial port details such as vendorId, productId and locationId... but those will be details about the UArt board (ie: FTDI chip) which is not specific to the sweep. This is a relatively involved process that requires querying each available port, just to make an educated "guess" about whether the device is more likely a sweep than other connected serial devices. 

What is the point of offering the simple version (that assumes portname) if we strongly recommend users avoid using it, and all our example content doesn't use it?  (Ie: if the novice user won't be using the most "simple" method, why offer it). I understand that we want to maintain an existing API. But there is very little existing code out there that makes use of the SDK. Seems like now is the perfect time to flesh out these choices in the API.

daniel-j-h:
Hrm the thing here is I'd like to keep the simple constructor for Linux since it will work most of the time. But not for Windows.

But I don't want to fragment the API, so I'm fine with you removing the simple version. Or better: changing it to take a dev path as only argument.

Let's do these changes sooner than later.

(Same for the c++impl and signal strength changes)

dcyoung:
> Hrm the thing here is I'd like to keep the simple constructor for Linux since it will work most of the time. But not for Windows.

I understand this. 

> But I don't want to fragment the API, so I'm fine with you removing the simple version. Or better: changing it to take a dev path as only argument.

Agreed. 

> Let's do these changes sooner than later.

Definitely agree. In this PR I tried to change the simple version to take a port name as the only argument. Did you want me to make a separate pull request for this?

dcyoung:
I think we should move forward with the portname addition to the simple constructor, but wait on the rest of this PR. 

I'd like to hold off until I finish working on Visual Studio support which should be our primary target on windows. I'm thinking of providing a `.sln` file and the visual studio project directories in the main `libsweep` directory. The src can stay put (in `libsweep/` or in the `src/` & `inc/` directories from PR #38 ) and the visual studio project's will simply reference source files by relative path from the `.sln` file. That way the separate build pipelines (makefile vs. VS) won't conflict, we won't have to modify the makefile pipeline at all, and we only have to include the src files in one location.

dcyoung:
There were too many conflicting goals (c++, file structure, win, visual studio etc), so I made a new PR #44 built off master, post C++ merge. We can ignore this old PR, but I figured I'd leave it here in case we want to reference the discussion.

PR #44, opened 2017-02-08
Windows Support

dcyoung:
Adds windows support
### Issues
- Changed the simple constructor to take a port name. This is a larger discussion. We could consider adding additional constructors to the API, but I think we should have the API require the port name. If we really want breakable constructors, we could include them in the hpp file or something and have them call the API's robust constructor.


dcyoung:
@daniel-j-h 
I pushed a change to master (bbbf132) and normally would have merged that with this `win` branch. You mentioned rebasing instead. I went ahead and rebased `win` off master,, but could not push without first pulling from the remote `win`. This seems to have cluttered the history more than a merge would have. Is there something I'm missing here? Would you prefer a forceful push rather than a pull & push?

libsweep/.gitignore
```diff
@@ -1,3 +1,288 @@
...
+# CodeRush
+.cr/
+
+# Python Tools for Visual Studio (PTVS)
+__pycache__/
+*.pyc
+
+# Cake - Uncomment if you are using it
+# tools/**
+# !tools/packages.config
```
daniel-j-h: Urgh do we really need all of those 300 loc for visual studio?

sweepjs/index.js
```diff
@@ -8,7 +8,7 @@ var util = require('util')
 if (require.main === module) {
   console.log('self-testing module');
 
-  var sweep = new sweepjs.Sweep();
+  var sweep = new sweepjs.Sweep("/dev/ttyUSB0");
```
daniel-j-h: Single quotes

sweepjs/README.md
```diff
@@ -14,7 +14,7 @@ See the [examples](examples) directory for an example streaming data from the de
 ### Interface
 
 ```javascript
-sweep = new Sweep();
+sweep = new Sweep("/dev/ttyUSB0");
```
daniel-j-h: Single quotes

libsweep/src/arch/win/serial_win.cc
```diff
@@ -0,0 +1,280 @@
+#include "serial.h"
+
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+#include <stdio.h>
```
daniel-j-h: We use C++ now so those should be cstdint etc. right?

libsweep/src/arch/win/serial_win.cc
```diff
@@ -0,0 +1,280 @@
...
+  SWEEP_ASSERT(port);
+  SWEEP_ASSERT(error);
+
+  for (int i = 0; i < 99; i++) {
+    char str[10];
+    sprintf(str, "COM%d", i);
+    if (strcmp(port, str) == 0) {
+      return i;
+    }
+  }
```
daniel-j-h: Hm this is a bit ugly. What about
- Checking for length > 3
- Calling `strtol` with [port + 3, last]
- Handling parsing errors

http://en.cppreference.com/w/cpp/string/byte/strtol

libsweep/src/arch/win/serial_win.cc
```diff
@@ -0,0 +1,280 @@
...
+
+device_s device_construct(const char* port, int32_t bitrate, error_s* error) {
+  SWEEP_ASSERT(port);
+  SWEEP_ASSERT(bitrate > 0);
+  SWEEP_ASSERT(error);
+
+  if (bitrate != 115200) {
+    *error = error_construct("baud rate is not supported");
+    return nullptr;
+  }
```
daniel-j-h: Hum why not? Is this a limitation of the serial implementation here?

The sweep device constructor provides a int32 bitrate argument - should this be removed then?

libsweep/src/arch/win/serial_win.cc
```diff
@@ -0,0 +1,280 @@
...
+
+  // read port number from the port name
+  error_s portNumError = nullptr;
```
daniel-j-h: I tended to avoid camelCase and used snake_case or single words where possible. Maybe stay in sync. with the rest of the code.

libsweep/src/arch/win/serial_win.cc
```diff
@@ -0,0 +1,280 @@
...
+  int portNum = detail_get_port_number(port, &portNumError);
+
+  if (portNumError || portNum < 0) {
```
daniel-j-h: Why checking < 0? Checking for error should be good enough, shouldn't it? Otherwise the interface is a bit weird.

libsweep/src/arch/win/serial_win.cc
```diff
@@ -0,0 +1,280 @@
...
+    *error = portNumError;
+    return nullptr;
+  }
+
+  TCHAR portName[32];
+  _stprintf_s(portName, sizeof(portName) / sizeof(TCHAR), _T("\\\\.\\COM%d"), portNum);
```
daniel-j-h: Are these Windows specific?

Can't you simply use

```c++
std::string fmt{"prefix "};
fmt += std::to_string(num);
fmt += " suffix";
```

libsweep/src/arch/win/serial_win.cc
```diff
@@ -0,0 +1,280 @@
...
+
+  // try to open the port
+  HANDLE hComm = CreateFile(portName,                     // port name
```
daniel-j-h: you can use `auto` here

libsweep/src/arch/win/serial_win.cc
```diff
@@ -0,0 +1,280 @@
...
+    return nullptr;
+  }
+
+  // specify timeouts (all values in milliseconds)
+  COMMTIMEOUTS timeouts;
+  if (!GetCommTimeouts(hComm, &timeouts)) {
+	  *error = error_construct("retrieving current serial port timeouts failed");
+	  CloseHandle(hComm);
+	  return nullptr;
+  }
```
daniel-j-h: Please clang-format

libsweep/src/arch/win/serial_win.cc
```diff
@@ -0,0 +1,280 @@
...
+
+  // try to close the overlapped read event
+  if (!CloseHandle(serial->ro_.hEvent)) {
+    ignore = error_construct("closing serial port overlapped read event failed");
+  }
+
+  // try to close the overlapped write event
+  if (!CloseHandle(serial->wo_.hEvent)) {
+    ignore = error_construct("closing serial port overlapped write event failed");
```
daniel-j-h: Hm no reason to set errors here we are not exposing anyway. You can just call your functions and hope for the best.

libsweep/src/arch/win/serial_win.cc
```diff
@@ -0,0 +1,280 @@
...
+  SWEEP_ASSERT(len >= 0);
+  SWEEP_ASSERT(error);
+
+  DWORD rx_read = 0;
+  DWORD totalNumBytesRead = 0;
+
+  // read bytes until "len" bytes have been read
+  while (totalNumBytesRead < (DWORD)len) {
+
+    if (!ReadFile(serial->fd, (char*)to + totalNumBytesRead, len - totalNumBytesRead, &rx_read, &(serial->ro_))) {
```
daniel-j-h: Probably has to be `unsigned char`, not sure what DWORD is

libsweep/src/arch/win/serial_win.cc
```diff
@@ -0,0 +1,280 @@
...
+      *error = error_construct("reading from serial device failed");
+      return;
+    }
+    totalNumBytesRead += rx_read;
```
daniel-j-h: Is this how to reliably read bytes on Windows? Can there be an interruption, as in EINTR on Linux?

libsweep/src/arch/win/serial_win.cc
```diff
@@ -0,0 +1,280 @@
...
+  }
+
+  SWEEP_ASSERT(totalNumBytesRead == (DWORD)len && "reliable read failed to read requested number of bytes");
```
daniel-j-h: Hm we should error here and not only assert. The callsite should not continue when we fail to read all requested bytes.

libsweep/src/arch/win/serial_win.cc
```diff
@@ -0,0 +1,280 @@
...
+  DWORD tx_written = 0;
+  DWORD totalNumBytesWritten = 0;
+
+  // clear the comm errors, and purge if need be
+  if (ClearCommError(serial->fd, &err, NULL) && err > 0) {
+    if (!PurgeComm(serial->fd, PURGE_TXABORT | PURGE_TXCLEAR)) {
+      *error = error_construct("purging tx buffer failed during seiral port write");
+      return;
+    }
+  }
```
daniel-j-h: How can comm errors happen? Does ClearCommError always write into err?

libsweep/src/arch/win/serial_win.cc
```diff
@@ -0,0 +1,280 @@
...
+
+  // write bytes until "len" bytes have been written
+  while (tx_written < (DWORD)len) {
+    if (!WriteFile(serial->fd,                               // Handle to the Serial port
+                   (const char*)from + totalNumBytesWritten, // Data to be written to the port
```
daniel-j-h: Probably also const unsigned char\* here

libsweep/src/arch/win/serial_win.cc
```diff
@@ -0,0 +1,280 @@
...
+                   len - totalNumBytesWritten,               // No of bytes to write
+                   &tx_written,                              // Bytes written
+                   &(serial->wo_))) {
+      *error = error_construct("writing to serial device failed");
+      return;
+    }
+    totalNumBytesWritten += tx_written;
+  }
+
+  SWEEP_ASSERT(((int32_t)totalNumBytesWritten == len) && "reliable write failed to write requested number of bytes");
```
daniel-j-h: Same here - we should error out

libsweep/src/arch/win/serial_win.cc
```diff
@@ -0,0 +1,280 @@
...
+}
+
+void device_flush(device_s serial, error_s* error) {
+  SWEEP_ASSERT(serial);
+  SWEEP_ASSERT(error);
+
+  // TODO: Before flushing/purging the port, should check for
+  // and handle or explicitly ignore any pending port erros
```
daniel-j-h: Needs to be done before we can merge this I guess?

libsweep/inc/protocol.h
```diff
@@ -111,18 +100,103 @@ typedef struct {
...
+  uint8_t cmdByte2;
+  uint8_t motor_speed[2];
+  uint8_t term;
+} response_info_motor_s;)
+#else
+#error "Only Clang, GCC and Visual supported at the moment, please open a ticket"
+#define SWEEP_API
+#define SWEEP_PACKED
+#endif
+#endif
```
daniel-j-h: Hm this got super ugly - we need to figure out a better way to handle these struct / packed issues..

libsweep/examples/viewer.cc
```diff
@@ -59,7 +64,7 @@ int main() try {
   thread.launch();
 
   // Now start scanning in the second thread, swapping in new points for every scan
-  sweep::sweep device;
+  sweep::sweep device(argv[1]);
```
daniel-j-h: `device{argv[1]}`

libsweep/examples/viewer.cc
```diff
@@ -24,7 +24,12 @@ static const sf::Color kColorDenim{80, 102, 127};
 using PointCloud = std::vector<sf::CircleShape>;
 using PointCloudMutex = std::mutex;
 
-int main() try {
+int main(int argc, char* argv[]) try {
+  if (argc != 2) {
+    std::cout << "Usage: ./example-viewer <portname>\n";
+    return 1;
```
daniel-j-h: include cstdlib
return EXIT_FAILURE

libsweep/examples/example.cc
```diff
@@ -5,8 +5,13 @@
 
 #include <sweep/sweep.hpp>
 
-int main() try {
-  sweep::sweep device;
+int main(int argc, char* argv[]) try {
+  if (argc != 2) {
+    std::cout << "Usage: ./example-c++ <portname>\n";
+    return 1;
```
daniel-j-h: include cstdlib
return EXIT_FAILURE

libsweep/examples/example.c
```diff
@@ -22,15 +22,22 @@ static void check(sweep_error_s error) {
     die(error);
 }
 
-int main() {
+int main(int argc, char* argv[]) {
+  if (argc != 2) {
+    fprintf(stdout, "Usage: ./example-c <portname>\n");
+    return 1;
```
daniel-j-h: include stdlib.h
return EXIT_FAILURE

libsweep/examples/README.md
```diff
@@ -19,12 +20,30 @@ To build:
...
-    ./example-c
-    ./example-c++
-    ./example-viewer
+    ./example-c /dev/ttyUSB0
+    ./example-c++ /dev/ttyUSB0
+    ./example-viewer /dev/ttyUSB0
+```
+
+
+#### To build on windows with MinGW-w64 and MSYS2:
```
daniel-j-h: Windows uppercase

libsweep/examples/README.md
```diff
@@ -19,12 +20,30 @@ To build:
...
+    cd build
+    cmake -G "MSYS Makefiles" ..
+    cmake --build .
+```
+**Note:** the viewer example is not compatible with windows currently.
+
+```bash
+    # run the examples (the number 5 is just an example, replace it with your COM port number)
+    ./example-c COM5
+    ./example-c++ COM5
```
daniel-j-h: I don't think you need the indentation when you block scopre via \```

libsweep/examples/README.md
```diff
@@ -2,12 +2,13 @@
...
 
-Requires `libsweep.so` to be installed.
 This can be either the dummy library always returning static point cloud data or the device library requiring the Scanse Sweep device to be plugged in.
 
 ### Quick Start
 
-To build: 
+#### To build on Linux: 
+
+Requires `libsweep.so` be installed, in `"/usr/lib"` for example.
```
daniel-j-h: I would remove the `/usr/lib` part. If not it would be more helpful for the user to know how to change linker and loader search paths.

libsweep/examples/CMakeLists.txt
```diff
@@ -18,9 +18,9 @@ target_include_directories(example-c SYSTEM PRIVATE ${LIBSWEEP_INCLUDE_DIR})
...
-pkg_check_modules(LIBSFML REQUIRED sfml-all)
+#include(FindPkgConfig)
+#pkg_check_modules(LIBSFML REQUIRED sfml-all)
 
-add_executable(example-viewer viewer.cc)
-target_link_libraries(example-viewer PRIVATE ${LIBSWEEP_LIBRARY} ${LIBSFML_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT})
-target_include_directories(example-viewer SYSTEM PRIVATE ${LIBSWEEP_INCLUDE_DIR} ${LIBSFML_INCLUDE_DIRS})
+#add_executable(example-viewer viewer.cc)
+#target_link_libraries(example-viewer PRIVATE ${LIBSWEEP_LIBRARY} ${LIBSFML_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT})
+#target_include_directories(example-viewer SYSTEM PRIVATE ${LIBSWEEP_INCLUDE_DIR} ${LIBSFML_INCLUDE_DIRS})
```
daniel-j-h: Rebase against master, I just merged a PR for running the examples on Travis which makes sfml2 builds optional.

libsweep/config.mk
```diff
@@ -1,15 +1,81 @@
...
+# Remove the dummy file from the src files
+SRC_FILES := $(filter-out $(SRC_DIR)/dummy.cc, $(SRC_FILES))
+
+# Specify src files for the dummy build (currently just dummy.cc)
+DUMMY_SRC_FILES := $(SRC_DIR)/dummy.cc
+
+# Then specify the obj files according to the structure of src directory
+# ie: (src/arch/unix/file_unix.cc -> obj/arch/unix/file_unix.o)
+OBJ_FILES := $(patsubst $(SRC_DIR)/%.cc,$(OBJ_DIR)/%.o,$(SRC_FILES))
+DUMMY_OBJ_FILES := $(patsubst $(SRC_DIR)/%.cc,$(OBJ_DIR)/%.o,$(DUMMY_SRC_FILES))
```
daniel-j-h: Hm this got a bit more complex than I'd like it to be. Do we now support VS but also MinGW on Windows?

libsweep/Makefile
```diff
@@ -1,26 +1,46 @@
...
+	@$(RM) $(INSTALL_DIR_INCLUDES)/sweep*
 
 watch:
-	while inotifywait -e modify *.cc *.h; do make; done
+	while ! inotifywait -e modify *.cc *.h; do make; done
 
 clean:
-	@$(RM) *.so *.o
+	@$(RM) *.so *.o *.dll
+	@$(RM) -rf bin obj
```
daniel-j-h: Should use vars from config now (?)

libsweep/config.mk
```diff
@@ -1,15 +1,81 @@
...
+# Set Options & Flags by Operating System
 ifeq ($(UNAME), Linux)
-  PREFIX ?= /usr
+  # For linux platforms
+  target = libsweep.so
+  dummy_target = dummy_linux
+  SRC_ARCH_DIR := src/arch/unix
+  INSTALL_DIR_LIB ?= /usr/lib
+  INSTALL_DIR_INCLUDES ?= /usr/include/sweep
+  LINKER = g++
```
daniel-j-h: Why is g++ the linker? ld should be fine?

g++ compiles, ld links and links in lstdlibc++ and lpthread

libsweep/src/arch/win/serial_win.cc
dcyoung: Currently communication protocol only supports 115200 baud rate. 

In the future we may support higher baud rates as the sensor performance or data formats require more. For this reason, I want to leave baud rate in the API. 

However, specifying the wrong baud rate can look like an unresponsive sensor, so to avoid confusion I report an unsupported baud rate. This is within the implementation, so we can easily change it later if we support different baud rates without breaking the API.

libsweep/config.mk
dcyoung: I ended up using g++ because it handled the `-mno-ms-bitfields` flag. Whereas ld threw an error:

> C:\msys32\mingw64\bin\ld.exe: unrecognised emulation mode: no-ms-bitfields

I'm not sure what the proper fix is here, but using g++ was working fine. I must have copied it over to the linux portion, I'm sure that could use ld without issue.

libsweep/config.mk
dcyoung: Yes, We support As it sits right now, things work for Visual Studio and MinGW-64 (with installation paths assuming MSYS2 installation).

libsweep/src/arch/win/serial_win.cc
dcyoung: The issue was the char set difference and it was causing a lot of trouble. I experimented with various methods to perform an MSDN UNICODE conversion, and ended up with the previous commit. Definitely not ideal.

Taking a second look at it today, I opted to change the Visual Studio project settings for libsweep to multi-byte char set. Which allowed me to use other methods, like the one you suggested.

libsweep/.gitignore
dcyoung: VS can certainly produce a lot of files, but I think this is a fairly standard gitignore for VS.

I'm sure for our use, it could be condensed, but is there any harm in a comprehensive gitignore?

dcyoung:
I attempted to make the VS support as non-invasive as possible. However, it's already a little different than a typical layout. Not only are all the source files outside of the projects, but the project directories are not at the same level as the main solution file. I attempted to counter this by collecting the project directories together under an explicitly labeled `VS_Projects` directory. I feel like changing this to `utils` or `scripts` is even more confusing. Given our initial software reports, windows will be the most common dev environment the first wave of sweep users. VS support is something we want to promote, not hideaway.

libsweep/.gitignore
daniel-j-h: No I was just surprised to see ~300 lines of files to ignore.

libsweep/config.mk
daniel-j-h: Hm the flag only affects compilation but not linking, right?
So it should probably be added to CXXFLAGS but not to LDFLAGS.

dcyoung: I was passing CXXFLAGS during the linking process as well. As i understood it, when specifying the `-shared` linker option it was safer to include the compilation options as well. I tried messing around a bit with getting ld to work, but my experimentation was introducing more complexity to the config and makefile. If you have a good fix for this on linux within the new config/makefile context I'm happy to test and adopt it for the windows options as well.

libsweep/src/arch/win/serial_win.cc
```diff
@@ -0,0 +1,291 @@
...
+  if (os_reader.hEvent == NULL) {
+    *error = error_construct("creating overlapped read event failed");
+    CloseHandle(hComm);
+    return nullptr;
+  }
+  // Create the overlapped write event. Must be closed before exiting
+  os_write.hEvent = CreateEvent(NULL, TRUE, FALSE, NULL);
+  if (os_write.hEvent == NULL) {
+    *error = error_construct("creating overlapped write event failed");
+    CloseHandle(hComm);
```
MikeGitb: Is this missing a `CloseHandle(os_reader.hEvent);` ?

MikeGitb:
Just a thought: Have you considered using cmake for building the library and the examples? That way you don't have to maintain separate makefiles and visual studio project files.

libsweep/src/arch/win/serial_win.cc
```diff
@@ -0,0 +1,291 @@
...
+  SWEEP_ASSERT(port);
+  SWEEP_ASSERT(error);
+
+  if (strlen(port) <= 3) {
+    *error = error_construct("invalid port name");
+    return -1;
+  }
+
+  char* end;
+  long parsed_int = std::strtoll(port + 3, &end, 10);
```
MikeGitb: shouldn't that line be

    long parsed_int = std::strtol(port + 3, nullptr, 10);

daniel-j-h: We strlen above anyway and can re-use the length to figure out the end here.

daniel-j-h:
Yeah the workflow should be roughly as follows:

git checkout -b myfeature
git add .
git commit
git push origin myfeature

To rebase your branch on top of latest master:

git rebase master
// optionally resolve conflicts
git push --force-with-lease origin myfeature

Never force push to other branches than your own and never to master.

daniel-j-h:
I see, okay. Just wanted to understand what's going on here.

daniel-j-h:
Hm not sure what you mean here; you compile with cxxflags and only pass
ldflags to the linker. But we can use g++ for linking, too. I think it
will change a few defaults, thouth (such as adding libstdc++ (?)).

daniel-j-h:
Yeah now with the Windows support coming in and the implementation
switching to C++11 anyway, we might want to move away from
Makefile+config.mk to a CMake based solution. The examples require CMake
already, too. From what I can see it would completely eliminate the need
for the VS_Projects directory.

libsweep/src/arch/win/serial_win.cc
dcyoung: I was under the impression that MSVC did not support strtol, but only strtoll as a replacement. I haven't tried passing it a nullptr, I'll test that out today.

dcyoung:
As far as I'm aware, cmake can produce VS solutions pretty easily. So I'd be happy to use a cmake based solution. Seems like a good way to unify everything.

dcyoung:
Will do. Would you like me to squash the commits in the process? or leave them as is?

daniel-j-h:
It's on you to squash or not for now :) we want to squash before we merge for sure.

dcyoung:
I still have to add windows specific bits, docs etc. So I'll let you know when its fine to merge

dcyoung:
I modified the Appveyor to:

- build libsweep
- build dummy
- install dummy
- build examples (windows only builds example-c and example-c++ currently)
- run examples

I have not yet added appveyor support for building or testing the bindings as I haven't looked into them on windows yet. At a glance, install/setup of the python bindings is erroring, and for the js bindings node-gyp fails on rebuild because sweep/sweep.h can't be found. 

I haven't looked into this any more yet, but the includes are definitely on the path, so its likely something else.

dcyoung:
I'm trying to get the node bindings building on windows and am hitting a few road blocks.

In the `bindings.gyp` file I had to manually specify paths for the includes and the library. The paths were on the environment variable `PATH` but that didn't seem to matter. How can we autodetect this? 

Once the bindings had a path for the library (sweep.lib) and the includes, node-gyp failed during the rebuild on unresolved symbols. I figured this might be something to do with the export/import of symbols... so I cleaned up the exportation of symbols (marked by SWEEP_API) when building the dll (makes use of flag BUILDING_DLL) and explicitly imported symbols when not building the dll. a84b6981c10d7df34a4462ff9863aecca8049b65

However, node-gyp still can't resolve the external symbols. Perhaps it cannot see the DLL? Or maybe I need to explicitly include the DLL somehow?? 

@daniel-j-h  Any tips ?

dcyoung:
A small fix to the python init script got things working. I included a sys module to detect the OS, not sure if that's best practice or not. I also went ahead added the python build and test to appveyor. 

Still stuck on exposing the DLL to node-gyp for the node bindings. See last post.

daniel-j-h:
Check the macOS Travis pull request for some cross-platform fixes. One of them is using

    ctypes.util.find_library('sweep')

to automatically switch between .so and .dylib, will probably work on Windows, too.

For your path and symbol problems I really can't help you much. I don't have experience with Windows, sorry :(

Maybe first try to manually build and link an example using libsweep on Windows. Then do

    npm install --verbose

and compare what's going on.

libsweep/README.md
```diff
@@ -21,7 +21,7 @@ For windows users, open a command prompt with administrative access:
 
     mkdir build
     cd build
-    cmake .. -DCMAKE_BUILD_TYPE=Release
+    cmake ..
```
daniel-j-h: Hm but on other system's a Release config is not the default.

We might not really care, but for Linux and macOS this decides what kind of optimization flags CMake passes to gcc.

libsweep/include/sweep/sweep.h
```diff
@@ -10,12 +10,28 @@
 extern "C" {
 #endif
 
+#if defined _WIN32 || defined __CYGWIN__ || defined __MINGW32__
+// If we are building a dll, set SWEEP_API to export symbols
+#ifdef BUILDING_DLL
```
daniel-j-h: What is the use-case for making it conditional on `BUILDING_DLL`? (Also please prefix with `SWEEP_`)

Check slide 35 here for what we basically want to mimic
http://www.slideshare.net/StefanusDuToit/cpp-con-2014-hourglass-interfaces-for-c-apis

libsweep/include/sweep/sweep.h
```diff
@@ -32,7 +48,7 @@ typedef struct sweep_scan* sweep_scan_s;
 SWEEP_API const char* sweep_error_message(sweep_error_s error);
 SWEEP_API void sweep_error_destruct(sweep_error_s error);
 
-SWEEP_API sweep_device_s sweep_device_construct_simple(sweep_error_s* error);
+SWEEP_API sweep_device_s sweep_device_construct_simple(const char* port, sweep_error_s* error);
```
daniel-j-h: :+1:

libsweep/src/win/serial.cc
```diff
@@ -0,0 +1,351 @@
...
+  SWEEP_ASSERT(error);
+
+  delete error;
+}
+
+static int32_t detail_get_port_number(const char* port, error_s* error) {
+  SWEEP_ASSERT(port);
+  SWEEP_ASSERT(error);
+
+  if (strlen(port) <= 3) {
```
daniel-j-h: If you save

    auto len = strlen(port);

here you can below do

    auto first = port + 3;
    auto last = port + len;
    auto v = strtoll(first, last);

libsweep/src/win/serial.cc
```diff
@@ -0,0 +1,351 @@
...
+  SWEEP_ASSERT(error);
+
+  if (bitrate != 115200) {
+    *error = error_construct("baud rate is not supported");
+    return nullptr;
+  }
+
+  // read port number from the port name
+  error_s port_num_error = nullptr;
+  int port_num = detail_get_port_number(port, &port_num_error);
```
daniel-j-h: const auto

libsweep/src/win/serial.cc
```diff
@@ -0,0 +1,351 @@
...
+
+device_s device_construct(const char* port, int32_t bitrate, error_s* error) {
+  SWEEP_ASSERT(port);
+  SWEEP_ASSERT(bitrate > 0);
+  SWEEP_ASSERT(error);
+
+  if (bitrate != 115200) {
+    *error = error_construct("baud rate is not supported");
+    return nullptr;
+  }
```
daniel-j-h: :+1: Good call on checking it here in the serial impl.

libsweep/src/win/serial.cc
```diff
@@ -0,0 +1,351 @@
...
+
+void device_read(device_s serial, void* to, int32_t len, error_s* error) {
+  SWEEP_ASSERT(serial);
+  SWEEP_ASSERT(to);
+  SWEEP_ASSERT(len >= 0);
+  SWEEP_ASSERT(error);
+
+  DWORD dw_num_read;
+  DWORD dw_num_to_read = (DWORD)len;
+  BOOL f_result = FALSE;
```
daniel-j-h: Is there a reason for BOOL and TRUE / FALSE? What about C++'s bool and true / false?

sweeppy/sweeppy/__init__.py
```diff
@@ -1,7 +1,8 @@
 import ctypes
+import ctypes.util
 import collections
 
-libsweep = ctypes.cdll.LoadLibrary('libsweep.so')
+libsweep = ctypes.cdll.LoadLibrary(ctypes.util.find_library('sweep'))
```
daniel-j-h: I think below the Python bindings are missing the simple constructor changes, here

https://github.com/scanse/sweep-sdk/blob/cdceb96751865ee2e632b0d229f8fe3ff00aff51/sweeppy/sweeppy/__init__.py#L19-L20

and here

https://github.com/scanse/sweep-sdk/blob/cdceb96751865ee2e632b0d229f8fe3ff00aff51/sweeppy/sweeppy/__init__.py#L83

libsweep/include/sweep/sweep.h
MikeGitb: Depending on whether you use the headers to create a dll or to consume a dll you need to prepend public functions with either `dllexport` or `dllimport`

libsweep/src/win/serial.cc
MikeGitb: @daniel-j-h: In case you aren't aware: The second parameter of `strtoll` is an **output**parameter of type `char**`. indicating the first character in the string that is not part of the parsed number. It is not a way to indicate the end in a string and if you don't care how many chars are parsed, you can just pass a `nullptr` (I think I mentioned this in the original windows PR)

daniel-j-h: Urgh you are right I misread - yes should be nullptr then

http://en.cppreference.com/w/cpp/string/byte/strtol

libsweep/include/sweep/sweep.h
MikeGitb: Isn't `BUILDING_DLL` the equivalent of `hairpoll_EXPORTS` on the slide you mentioned?

libsweep/README.md
```diff
@@ -4,19 +4,36 @@ Low-level Scanse Sweep LiDAR library. Comes as C99 library `sweep.h` with option
...
 This dummy library is API and ABI compatible. Once your device arrives switch out the `libsweep.so` shared library and you're good to go.
 
+For windows users, open a command prompt with administrative access:
+
+```bash
+mkdir build
+cd build
+# The 14 stands for VS 2015
+cmake .. -G "Visual Studio 14 Win64"
+cmake --build . --config Release
```
dcyoung: DCMAKE_BUILD_TYPE is specific to make generators (or at least it doesn't work with IDE style generators (Visual Studio, and potentially XCode).  I do believe that `--config Release` won't cause any issues with make style generators... if this can be verified maybe we should switch over to that for unix as well, just to be consistent?

libsweep/include/sweep/sweep.h
dcyoung: @MikeGitb that is correct. 
@daniel-j-h I switched the naming convention to more closely match the slides you posted, hopefully that clarifies any confusion.

libsweep/src/win/serial.cc
dcyoung: I had originally been saving the end value in case we wanted to add a check for bad inputs with spaces, ie: "COM5 6". But I think what we have is sufficient. I switched it to use `nullptr`.

dcyoung:
Ok a few things:

1. Building sweepjs requires a 64 bit version of the sweep dll for some reason. I'm not sure how to get around this. For now I documented it, and all our README's show examples of building a 64bit version.

2. The process of building sweepjs was more involved than I had hoped (ie: requires user to manually rename file). I documented it, but its still an extra step. Perhaps this could be performed in a post-install script?

3. The sweepjs bindings are currently using hardcoded paths for the include and lib installation dirs. Although these are likely similar among most win users, I don't like that it can't auto find them. For now I included a comment in the docs about changing the paths for abnormal installation directories.

4. The sweepjs `scan` method did not work because the async scan worker's `HandleOKCallback` method was never called, despite `execute` being called successfully. Somehow the worker was performing the callback, but from where I don't know. It wasn't an error, and the callback did not include any arguments (not even an err arg). During my debugging, I added another test method that also implements an asynchronous queue worker and for some reason the `scan` method started working properly, without ever invoking my added method. I have NO IDEA why this works, and I am not comfortable relying on it as the fix. But I included it in the commit to see if anyone knows why it would fix the behavior of the `scan` method.

5. `index.js` assumes a specific port on linux. I changed `index.js` to accept an input arg and added the specific port example as a passed argument in the `test` script. I changed the windows docs to instruct the user to call `node index COMX` instead of using `npm test`. 

6. The suggested commands (in the README) build a 64bit version of the dll. This was mainly sparked by node-gyp bindings requiring a 64bit version. However, witnessing the appveyor fail, I just discovered that the default python installation is 32 bit and therefore can't load the 64bit dll. I could specify instructions in the `sweeppy/README` that say you have to build a version of the library that matches your python installation. I could then change the appveyor to build and install the 32 bit version before running the python test. For now I removed the python test from appveyor until this is sorted.

daniel-j-h:
Urgh sorry for the troubles with NodeJs bindings on Windows. What do you think of just removing Windows support for NodeJs bindings altogether for now. If users want this they have to help out here.

Maybe put a note into the readme - NodeJs on Windows is probably not a popular setup anyway.

Can you rebase on master and resolve merge conflicts please.

MikeGitb:
Maybe this is selfish, but I'd also say that it would be preferable to merge the core part now and then add missing functionality in incremental PRs instead of trying to first get the windows support to full feature parity while simultaneous having to track all changes to the master branch. As this PR touches a lot of files - including the build system - you'll get a merge conflict pretty much every time a non-trivial change is made to the master branch.

dcyoung:
I just pushed some changes that:

- noted in the docs that the architecture of the library and python must match (ie: for an installed 64bit DLL, user must be running python 64bit). 
- adjusted the bindings to account for the simple constructor change @daniel-j-h  please confirm this was done properly
- added a command line value for the portname

This has been tested on my own system, but I still need to add python tests back to the appveyor once i figure out how to specify python architecture or find a way to uninstall and reinstall different versions of the library between testing node and python.

daniel-j-h:
I adapted all examples to the device constructor and tested it locally.
I also removes the dummy function for Windows in the Node bindings.

We should wait on Travis. Then this is ready to merge :sparkles:

dcyoung:
@daniel-j-h  Thanks for the updates. However removing the dummy function in the Node bindings breaks the scan function on windows, even though they are completely separate. See my earlier comments about it. I don't know what the proper fix here is as I'm completely baffled the dummy function resolved the problem.

daniel-j-h:
Yes as I said, Node bindings not working on Windows is better than having a mysterious function in there no one knows why it makes things work. Is users need Node bindings on Windows they have to help out here.

Side note: macOS builds on Travis are quite slow, we have to wait for them.

dcyoung:
I agree it's not a fix. But currently the docs indicate the nodejs bindings work on windows. If you remove the hacky fix, we should update those docs as well. 

Towards finding a proper solution, is there anything in the node bindings (related to the instantiation of the worker) that might be compiler specific? Most of the examples I looked through for the using the async workers defined a worker class elsewhere in the code, rather than within the function. Is it possible this difference is related to the issue?

dcyoung:
@daniel-j-h  I just added a commit that fixes the issue on windows. I'm not sure why the original version was failing, but moving it to a class outside the `scan` method resolved the issue.

dcyoung:
It appears the binding file permits the specification of multiple possible locations for the library and includes. I added support for the default installation directory on both x86 and x64. 

`npm install` now works on both x64 and x86 installations assuming the node-gyp architecture matches the library. 

I switched the appveyor to build and install a x86 version of the dummy library, so it could test both Sweepjs and Python bindings. Both of which are now building and testing successfully.  Hooray!

daniel-j-h:
Interesting, I have no idea how the inline type can be a problem, but if this fixes the building issues :+1: 

Seems like the macOS builds on Travis were erroring out since the examples' CMakeLists had an early exit built into them for macOS. Let's wait for it and then finally merge this.

daniel-j-h:
Merged. Thank you for this major push for Windows support!

Issue #27, opened 2017-01-24
Parse DX receipt in `sweep_device_stop_scanning`, instead of sending DX command twice.


dcyoung:
Sending a `DX` command  (host->sensor) to a streaming sensor stops data acquisition. However, this doesn't happen instantaneously. Once the `DX` command is sent out, the host will still receive a few `Data Block` receipts before seeing the `DX` receipt.

Currently the `sweep_device_stop_scanning` uses the following technique:

1. Send `DX` command (host -> sensor)
2. Wait a while (5 seconds), for any remaining `Data Block` receipts and the `DX` receipt to arrive (sensor -> host).
3. Flush the serial port.
4. Send another `DX` command (host -> sensor)
5. Read the `DX` receipt (sensor -> host) that returns as a response to step 4

This was really a temporary workaround and should be replaced by an actual check for the `DX` receipt among the incoming `Data Block` receipts. The initial reaction might be to assume we cannot reliably check for a `DX` receipt among incoming `Data Block` receipts because the bytes of a `Data Block` are variable. The fear is that they could theoretically take the appearance of a `DX` receipt. However, we can still reliably discern the two because:
- While it is true that `Data Block` bytes 1-6 can take any form, byte 0 (sync/error byte) is limited to very specific error codes and is therefore controllable. We can guarantee that byte 0 will never look like the first byte of a `DX` receipt. 
- The sensor guarantees the transmission of complete `Data Block` receipts before sending a `DX` receipt. That is to say that a `DX` receipt will never come mid-`Data Block`.
- The sensor guarantees that no `Data Block` receipts are transmitted after the `DX` receipt is transmitted.

Obviously we can't protect against data corruption with 100% reliability, but we can certainly protect against valid `Data Block` receipts being misinterpreted as a `DX` receipt.

In `sweep_device_stop_scanning`, we should do something along the lines of:
1. Send a `DX` command (host -> sensor)
2. Look at the incoming bytes, and discern if they belong to a `Data Block` receipt or a `DX` receipt. 
3. In the case of a `Data Block` receipt, handle it... (as we would during scanning, or simply trash it). In the case of a  `DX` receipt, we proceed as if we picked up from step 5 above.

dcyoung:
Looks like the current implementation (sleep) is causing an issue. 

The example-c++ and example-viewer are throwing an occasional (read ~10% of the time) error on program exit.

> unable to receive stop scanning command response

[...]

PR #45, opened 2017-02-11
Implements Communication Error Bit, resolves #20

### LINKED ISSUE #20: Implement error code bit
https://github.com/scanse/sweep-sdk/commit/a0d54589daa4cd717b917376a6c2deeff96a143f introduced an error code bit. Currently we only check the sync bit [here](https://github.com/scanse/sweep-sdk/blob/a0d54589daa4cd717b917376a6c2deeff96a143f/libsweep/sweep.c#L200) to check when a new full scan starts. If the error bit is set we should throw away the scan.

daniel-j-h:
For #20.

@kent-williams @dcyoung can it happen that the sync bit _and_ the error bit is set? (or really: multiple bits, not specific to those two) Also any insights to error codes for other bits? Or are they unused on the firmware level, too, at the moment?

### DIFF
--- a/libsweep/protocol.h
+++ b/libsweep/protocol.h
@@ -75,14 +75,29 @@ typedef struct {
 static_assert(sizeof(response_param_s) == 9, "response param size mismatch");
 
 typedef struct {
-  uint8_t sync_error;
-  uint16_t angle; // see u16_to_f32
+  uint8_t sync_error; // see response_scan_packet_sync::bits below
+  uint16_t angle;     // see u16_to_f32
   uint16_t distance;
   uint8_t signal_strength;
   uint8_t checksum;
 } SWEEP_PACKED response_scan_packet_s;
 
 static_assert(sizeof(response_scan_packet_s) == 7, "response scan packet size mismatch");
+
+namespace response_scan_packet_sync {
+enum bits : uint8_t {
+  sync = 1 << 0,                // beginning of new full scan
+  communication_error = 1 << 1, // communication error
+
+  // Reserved for future error bits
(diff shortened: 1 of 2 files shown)


dcyoung:
You can definitely have a sync bit AND error bits set on a single DATA_BLOCK. 

See the last post  #27 for a better description of the error codes. There is at least one error code value we may have to avoid producing, so instead of relying on individual error bits to separately indicate different issues, we may want to just consider bits `e0-6` as a single 7bit integer error code, of which we avoid using 1 particular error code value. This will also let us simplify the checking to a single error code:

```c
    // Consider the DATA_BLOCK contents...
    // check that the error bits (e0-6) are all 0 (ie: DATA_BLOCK contains NO ERRORS)
    if ( !(responses[received].sync_error >> 1) ) {

      // Check if DATA_BLOCK is a sync reading (ie: first reading of a new 360deg scan)
      // & if we've already been gathering DATA_BLOCKs
      if ( responses[received].sync_error == 1 && received > 0) {
        // Note the end of the scan
        last = received;
        //FIXME: add check that "last" does not exceed SWEEP_MAX_SAMPLES
        
        // BREAK to return the scan (NOT including this DATA_BLOCK)
        break;
      }
    }
    else {
      // some error
      //FIXME: Handle each type of error code and shutdown gracefully if need be. 
      switch(responses[received].sync_error >> 1 ) {
          case 1: // communication error
          case ....
       }
```

daniel-j-h:
Thanks - I somehow forgot about this. Changed it to an error check by shifting out the sync bit.

Issue #19, opened 2016-11-12
High-resolution timestamp for each sample


daniel-j-h:
At the moment we accumulate a full 360 degree scan and return this scan to the user.

Many use-cases need high-resolution synchronization which is currently not possible on a per-sample basis to do for the user (could be estimated using motor speed and timestamp on a scan, though).

For usability reasons it would be great to have a timestamp per angle, distance sample.

Tasks:
- [ ] time.h needs a timestamp interface
- [ ] time_unix.c implementation
- [ ] sweep.h needs a timestamp interface for samples (similar to get_distance, get_angle, ..)
- [ ] sweep.c and dummy.c implementations
- [ ] add to sweep.hpp
- [ ] add to sweeppy
- [ ] add to sweepjs

kent-williams:
Just FYI, We plan to add a timestamp to each scan packet in the future, but it first requires increasing the serial baud rate to make room for the extra bytes. I think this is a much needed addition though for the time being.

MikeGitb:
Will those time stamps be available in the next batch of devices you ship?

MikeGitb:
Is attaching a time stamp to each measurement package still of interest? If so, I could provide a PR.

PR #46, opened 2017-02-11
Implements Sample Rate, resolves #21

### LINKED ISSUE #21: Implement sample rate
https://github.com/scanse/sweep-sdk/commit/449565f45fea61a7b99adc942d4f6c4ecb7005b4 introduced commands to get but more importantly set the sample rate.

For https://github.com/scanse/sweep-sdk/issues/7 the sample rate was left out since we could only read it. Now that we can modify it we should implement it for users to be able to set it.

daniel-j-h:
500 Hz:

500hz

750 Hz:

750hz

Looks like the signal strength goes down quite a bit from my first experiences. No real difference between 750 Hz and 1000 Hz. Resolution goes up (even though not _that_ much (?)), looks good over all!

@dcyoung @kent-williams can you take a quick look. Especially since you know more about firmware plans. I opted for the user having to explicitly specify the Hz, and let the README document the support for 500 Hz, 750 Hz and 1000 Hz.

From your docs it looks like the device can be off by ~50-100 Hz. Which means high-precision applications probably require accurate timestamps (?) (see https://github.com/scanse/sweep-sdk/issues/19).

References:
- https://github.com/scanse/sweep-sdk/commit/449565f45fea61a7b99adc942d4f6c4ecb7005b4
- https://github.com/scanse/sweep-sdk/commit/83043a66d2ca221c7c55b902e5be5781e839dd73

### DIFF
--- a/libsweep/README.md
+++ b/libsweep/README.md
@@ -168,6 +168,22 @@ The device supports speeds of 1 Hz to 10 Hz.
 In case of error a `sweep_error_s` will be written into `error`.
 
 ```c++
+int32_t sweep_device_get_sample_rate(sweep_device_s device, sweep_error_s* error)
+```
+
+Returns the `sweep_device_s`'s sample rate in Hz.
+In case of error a `sweep_error_s` will be written into `error`.
+
+```c++
+void sweep_device_set_sample_rate(sweep_device_s device, int32_t hz, sweep_error_s* error)
+```
+
+Sets the `sweep_device_s`'s sample rate in Hz.
+The device supports sample rates of 500 Hz, 750 Hz and 1000 Hz.
+The device guarantees for those sample rates but they can be slightly higher by a maximum of roughly 50-100 Hz.
+In case of error a `sweep_error_s` will be written into `error`.
+
+```c++
 void sweep_device_reset(sweep_device_s device, sweep_error_s* error)
 ```
 
--- a/libsweep/dummy.cc
+++ b/libsweep/dummy.cc
@@ -13,6 +13,7 @@ typedef struct sweep_error {
 typedef struct sweep_device {
   bool scanning;
   int32_t motor_speed;
+  int32_t sample_rate;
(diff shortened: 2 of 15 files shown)


dcyoung:
Hmm, sample rate code `03` (~1000Hz) should be noticeably different. Were you able to verify that the device was returning code `03` after setting it?

In general, I'm wondering if we should avoid having the user specify a value in HZ. If we eventually add support for specifying more specific sample rates, this could be of use... but I'm not sure that is even possible @kent-williams.  

Currently the 3 supported sample rate codes (`01`, `02`, `03`), correspond more to a target range than a specific value. It might be better if we allow the user to specify a sample rate code which we define clearly in the protocol and API as targeting a certain sample rate range. This way people are not inclined to misinterpret the parameter as a true variable. Otherwise, we should specify a sample rate enum or mapping of code to HZ value. Making structures or queries like this available through the API will allow people to build application logic around the codes and values. Then we can update the values of codes without breaking user's code.

daniel-j-h:
Yes I made sure to check the get sample rate response. You can try it yourself, e.g. add a the set function call to one of the examples and run it. Then run the viewer.

Re. the API: we could add enums for those codes for sure, as in

```c++
enum {
  500_600_HZ = 1,
  ..
};
```

I'd like to hear your plans for future firmware updates / devices on this, though.
The enum would need a bit of wrapping for the bindings, too.

daniel-j-h:
Here are some more comparisons from today (click to enlarge).

With motor speed 3Hz. Sample rate from left to right: 500Hz, 750Hz, 1000Hz:

Motor 3Hz, Sample Rates


With motor speed 7Hz. Sample rate from left to right: 500Hz, 750Hz, 1000Hz:

Motor 7Hz, Sample Rates

Do these observations look reasonable from your experiences @kent-williams @dcyoung? Not sure if we should recommend users leaving the sample rate at 500Hz except they know what they're doing.

Issue #13, opened 2016-11-02
What layer could we introduce an IP socket transport?


kent-williams:
Several applications are going to require communicating with sweep over IP. What layer of the SDK would be most appropriate to introduce this functionality?

daniel-j-h:
In https://github.com/scanse/sweep-sdk/issues/47 I added an example using ZeroMQ and Protobuf for a pub/sub setup. Is this good enough or should we bring this out of the examples and into the main `libsweep` project?

What are the requirements here, what do we actually need?

daniel-j-h:
Closing as not actionable on our side. Examples are there, we could install the publisher and subscriber, in addition to the proto scheme if needs be.

Issue #12, opened 2016-11-01
More graphical and fancy demos


daniel-j-h:
At the moment I only started to play with the point-cloud 3d map idea, but we don't have much examples that are more beautiful than "just" printing some numbers on the screen.

I understand having library + bindings is a great first step, but users want fancy examples and use-cases.

daniel-j-h:
For the record here is the current viewer and a density based clustering demo:

![sebshli](https://cloud.githubusercontent.com/assets/527241/20300444/92ade432-ab1f-11e6-9d96-a585df3fe471.png)

![dbscan](https://cloud.githubusercontent.com/assets/527241/20300478/b5ae968e-ab1f-11e6-8ee0-d24aedd835f9.png)

PR #47, opened 2017-02-11
Adds Protobuf + ZeroMQ Networking Example

daniel-j-h:
Simple Pub-Sub example:
- Publisher polls scans from device, encodes them as Protobuf messages, sends them out.
- Subscribers connect to publisher, decode messages, print to stdout.

I wanted to see how hard it would be to pull this off and at the same time provide additional examples. Maybe it helps.

See #12, #13.

References:
- http://api.zeromq.org/2-1:zmq-cpp

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,30 +11,22 @@ matrix:
       addons:
         apt:
           sources: ['ubuntu-toolchain-r-test']
-          packages: ['g++-6', 'cmake', 'python3', 'nodejs', 'npm', 'libsfml-dev']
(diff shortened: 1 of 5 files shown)


daniel-j-h:
Merging this. @kent-williams @dcyoung do we still need to think about https://github.com/scanse/sweep-sdk/issues/13 or can this example be a starting point for users? Do we have to support a networking solution out of the box? (If not please close the issue and reference this Protobuf + ZeroMQ example

Issue #9, opened 2016-11-01
Refactor Error Translation


daniel-j-h:
The [serial sub-system](https://github.com/scanse/sweep-sdk/blob/f7fbfbd2e304887a23d13678b40e64ecff770485/libsweep/serial.h) provides it's own error types to be fully encapsulated and to not go near circular dependencies. The translation to [sweep's error types](https://github.com/scanse/sweep-sdk/blob/master/libsweep/sweep.h) happens manually all over the place in [sweep.c](https://github.com/scanse/sweep-sdk/blob/f7fbfbd2e304887a23d13678b40e64ecff770485/libsweep/sweep.c) (e.g. https://github.com/scanse/sweep-sdk/blob/f7fbfbd2e304887a23d13678b40e64ecff770485/libsweep/sweep.c#L69-L73 where the serial error message is even discarded).

Refactor this translation into a helper function.

daniel-j-h:
Update: we're using C++11 in our implementation now and only expose error codes to the outside (C API boundary). When working on this think about if it would make sense to throw sub-system exceptions catching them at the boundary. See discussion in https://github.com/scanse/sweep-sdk/pull/33#issuecomment-277514679.

PR #48, opened 2017-02-12
Pr/exceptions

MikeGitb:
This is more a prove of concept than a real PR, but it demonstrates how exceptions could be used to simplify the code as discussed in issue #9.

Unfortunately, this is a quite pervasive change and additionally intertwined with a few other changes I made on the way (see comment on last commit). So I don't know if merging it is realistic but it should give you an impression of what it could look like.

In any case, please note that a solution like this will not produce the same error messages as before, because I omitted the translation step when an error bubbles up. You could do this with exceptions by adding a try, catch and rethrow block everywhere but that would essentially eliminate the advantage of using exceptions in the first place.

### DIFF
--- a/libsweep/CMakeLists.txt
+++ b/libsweep/CMakeLists.txt
@@ -1,0 +1,19 @@
+project(sweep)
+
+cmake_minimum_required(VERSION 3.5)
+
+SET(COMMON_SRC_FILES sweep.cc protocol.cc sweep.h sweep.hpp serial.h protocol.h)
+
+if (MSVC)
+    SET ( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /W4 /WX" )
+    SET (PALTFORM_SRC_FILES serial_win.cc)
+else ()
+    SET ( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra -Werror" )
+	SET (PALTFORM_SRC_FILESserial_unix.cc)
+endif()
+
+set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS true)
+
+
+add_library(${PROJECT_NAME} SHARED ${COMMON_SRC_FILES} ${PALTFORM_SRC_FILES})
+
--- a/libsweep/Error.h
+++ b/libsweep/Error.h
@@ -1,0 +1,13 @@
+#pragma once
+#include <stdexcept>
+
+namespace sweep {
+struct ErrorBase : std::exception {
+  // parameter MUST be a STRING LITTERAL
+  template <size_t N> ErrorBase(const char (&msg)[N]) : _msg{&msg[0]} {}
+  const char* what() const noexcept override { return _msg; };
+
+private:
+  const char* _msg;
+};
+}
--- a/libsweep/protocol.h
+++ b/libsweep/protocol.h
@@ -8,16 +8,16 @@
 
 #include <stdint.h>
 
+#include "Error.h"
 #include "serial.h"
 #include "sweep.h"
 
 namespace sweep {
 namespace protocol {
 
-typedef struct error* error_s;
-
-const char* error_message(error_s error);
-void error_destruct(error_s error);
+struct Error : ErrorBase {
+  using ErrorBase::ErrorBase;
+};
 
 // Command Symbols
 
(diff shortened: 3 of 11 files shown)


libsweep/protocol.h
```diff
@@ -147,8 +151,8 @@ inline void speed_to_ascii_bytes(int32_t speed, uint8_t bytes[2]) {
   // Speed values are still ASCII codes, numbers begin at code point 48
   const uint8_t ASCIINumberBlockOffset = 48;
 
-  uint8_t num1 = (speed / 10) + ASCIINumberBlockOffset;
-  uint8_t num2 = (speed % 10) + ASCIINumberBlockOffset;
+  uint8_t num1 = (uint8_t)(speed / 10) + ASCIINumberBlockOffset;
+  uint8_t num2 = (uint8_t)(speed % 10) + ASCIINumberBlockOffset;
```
daniel-j-h: Why the cast now?

libsweep/serial.h
```diff
@@ -7,24 +7,29 @@
...
-void device_write(device_s serial, const void* from, int32_t len, error_s* error);
-void device_flush(device_s serial, error_s* error);
+struct Error : public ErrorBase {
+	using ErrorBase::ErrorBase;
+};
+
+class Device {
+public:
+	Device(const char* port, int32_t bitrate);
+	Device(Device&& other) = default;
```
daniel-j-h: Why a move ctor but not a move assignment operator? Also why not noexcept?

Let's go a step back here: why the unique_ptr state below anyway?

libsweep/sweep.cc
```diff
@@ -42,10 +22,34 @@ const char* sweep_error_message(sweep_error_s error) {
...
+    *error = sweep_error_construct(e.what());
+  } catch (const std::bad_alloc&) {
+    *error = sweep_error_construct("Could not allocate enough memory");
+  } catch (...) {
+    *error = sweep_error_construct("Unknown exception");
+  }
+  return {};
+}
+
+template <class F> void translateExceptionV(F&& f, sweep_error_s* error) {
```
daniel-j-h: Hm what about having two overloads one for void and one for a return value instead of the V suffix?

libsweep/sweep_device.hpp
```diff
@@ -0,0 +1,145 @@
...
+
+#include "protocol.h"
+#include "serial.h"
+
+#include <chrono>
+#include <string>
+#include <thread>
+
+#include "sweep.h"
+#include "sweep_device.hpp"
```
daniel-j-h: Self include?

libsweep/sweep_device.hpp
```diff
@@ -0,0 +1,145 @@
+#pragma once
```
daniel-j-h: Can you be consistent with all other header guards please.

libsweep/sweep_device.hpp
```diff
@@ -0,0 +1,145 @@
...
+
+  sweep_scan_s get_scan();
+
+  int32_t get_motor_speed();
+  void set_motor_speed(int32_t hz);
+
+  void reset();
+
+private:
+  sweep::serial::Device serial; // serial port communication
```
daniel-j-h: The public API now directly depends on internals. We want to avoid that.

libsweep/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+cmake_minimum_required(VERSION 3.5)
+
+SET(COMMON_SRC_FILES sweep.cc protocol.cc sweep.h sweep.hpp serial.h protocol.h)
+
+if (MSVC)
+    SET ( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /W4 /WX" )
+    SET (PALTFORM_SRC_FILES serial_win.cc)
+else ()
+    SET ( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra -Werror" )
+	SET (PALTFORM_SRC_FILESserial_unix.cc)
```
daniel-j-h: Space missing?

libsweep/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+project(sweep)
+
+cmake_minimum_required(VERSION 3.5)
```
daniel-j-h: Ubuntu 14.04 has CMake 2.8.12. and I'd like to make it as painless as possible at least going back to Trusty.

MikeGitb:
As I said, a prove of concept. I can compose a proper PR incorporating your feedback, but would rather wait until the VS-Support is integrated and base it on that. Otherwise feel free to just copy/use whatever parts you deem usefull.

libsweep/CMakeLists.txt
MikeGitb: Most likely you need a completely different cmake file for your project (e.g. it doesn't have an install target and I haven't even tested if this works correctly with the examples). It was just a simple way for me to use Visual Studio - I probably shouldn't have checked it in in the first place.

libsweep/sweep_device.hpp
MikeGitb: Sorry for the confusion: `sweep_device` is not meant to be part part of the public API. The public c++ API is still `Sweep` and the public C-api still consists of the functions in `sweep.h` and the opaque pointer. Admittedly I still think that it is unnecessary to provide a stable ABI for c++ projects, but that is not my decision and

libsweep/sweep.cc
MikeGitb: Yes, that was my first intention but I couldn't think of a simple SFINAE expression from the top of my head, so I went with the *manual* solution

libsweep/serial.h
MikeGitb: I probably just forgot the move assignment operator. The pimple idiom was used to keep the header free of platform specific code (and in particular includes) just like the original. but other than that there is no reason why the state couldn't be stored in normal member variables (guarded by an appropriated `#ifdef`).

libsweep/protocol.h
MikeGitb: Because VS gives a warning here due to the narrowing conversion. But they should actually be `static_cast`s instead of c-style casts.

libsweep/Error.h
```diff
@@ -0,0 +1,13 @@
...
+
+namespace sweep {
+struct ErrorBase : std::exception {
+  // parameter MUST be a STRING LITTERAL
+  template <size_t N> ErrorBase(const char (&msg)[N]) : _msg{&msg[0]} {}
+  const char* what() const noexcept override { return _msg; };
+
+private:
+  const char* _msg;
+};
```
MikeGitb: Yes, I pretty much copied the currrent error mechanism as I wasn't sure if the error messages being string litterals was a desired property or an undesired limitation of the current implementation.

PR #51, opened 2017-02-14
Add setuputils script for py bindings

kenzierocks:
Python packages usually have a `setup.py` included to install the scripts.


daniel-j-h:
Thanks! It's a good idea to make this properly installable!

I think we want to fill in a bit more details, e.g. license, use README.md as a readme, etc.

Could you adapt `.travis.yml` to install the Python module. Then break of the test script from the library

https://github.com/scanse/sweep-sdk/blob/6c3ae36af4a1ae6cb5dee1444850728a96dc4497/sweeppy/sweeppy.py#L215-L230

into a small `example.py` `import sweeppy` (without any sys.path tricks :P) and run the test routines.

daniel-j-h:
Here's a bit more context for what we want to do in the future: eventually we want to package the pre-built shared library `libsweep.so` for the Python module (as a wheel). This would allow users to `pip install sweep` and have everything working magically. Here are some good readings:

- [Setuptools](https://setuptools.readthedocs.io/en/latest/setuptools.html)
- [Packaging and Distributing Projects](https://packaging.python.org/distributing/) - section on Wheels
- [Python on Wheels](http://lucumr.pocoo.org/2014/1/27/python-on-wheels/)

Building these platform dependant libraries is a bit more involved, though. The common way to do it for Linux is to spawn up a Docker container with an old libc, build the library there (for compatibility), and distribute it. For macOS and Windows it may be a bit easier.

Anyway, that's just to give some context of our future plans.
For now a Setuptools based installation and requiring the user to compile `libsweep.so` is fine.

kenzierocks:
It sounds like PyPI doesn't allow Linux wheels right now, so I don't think there's a way to include a compiled `libsweep.so`. I've added everything else, though.

daniel-j-h:
Hm not sure where you got the information from that PyPI does not support Linux wheels. I think what we have to do is build the shared library on a "old enough" system (CentOS 5) and restrict us a bit for the wheels to be part of the `manylinux1` ecosystem, see:

- https://www.python.org/dev/peps/pep-0513/
- https://github.com/pypa/manylinux

These wheels work happily on basically all Linux systems currently out there. At least the ones we care about having binary builds for. It would be perfect if we could let Travis CI build those wheels in a Docker container for master.

I'd like to split the wheels discussion off of your pull request, though. If that is fine with you. Can you squash and rebase your branch, then I'd like to merge it into master.

kenzierocks:
It's squashed and rebased on `master`.

daniel-j-h:
Thank you! Happily merging now! :)

PR #52, opened 2017-02-14
Switches libsweep from Makefile to CMake, resolves #50

### LINKED ISSUE #50: Switch libsweep from Makefile to CMake
We're currently using a `Makefile` + `config.mk` for the build system. Now with the implementation switching to C++11 and @dcyoung polishing up the Windows branch we should re-visit this decision.

- [Makefile](https://github.com/scanse/sweep-sdk/blob/6c3ae36af4a1ae6cb5dee1444850728a96dc4497/libsweep/Makefile)
- [config.mk](https://github.com/scanse/sweep-sdk/blob/6c3ae36af4a1ae6cb5dee1444850728a96dc4497/libsweep/config.mk)

The examples are using CMake, too, and we provide a [`FindSweep.cmake`](https://github.com/scanse/sweep-sdk/blob/master/libsweep/cmake/FindSweep.cmake) CMake module already.

[Here](https://github.com/scanse/sweep-sdk/blob/6c3ae36af4a1ae6cb5dee1444850728a96dc4497/libsweep/examples/CMakeLists.txt#L7-L9) is how users can then include `libsweep` into their CMake based build system.
Related issue: [install FindSweep.cmake on the user's system](https://github.com/scanse/sweep-sdk/issues/17).

Should probably land before merging the Windows support branch to eliminate the need for Visual Studio specific project files, if we decide on going this route.

Extra requirements for a CMake based build system:
- Require CMake 2.8.12, so we can build without having to upgrade CMake going back to [Trusty](http://packages.ubuntu.com/trusty/cmake)
- Add SONAME (or equivalent) to shared library in a portable way
- Provide install target for `libsweep.so`+`sweep/sweep.h`+`sweep/sweep.hpp`

daniel-j-h:
Would love to get this properly tested and then merged in before we merge the full Windows integration.

Does this CMake integration look reasonable to you?


libsweep/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
+cmake_minimum_required(VERSION 2.8.11)
+project(sweep LANGUAGES C CXX)
+
```
MikeGitb: Don't know why, but this triggers 

> CMake Error: Error required internal CMake variable not set, cmake may be not be built correctly.
Missing variable is:
CMAKE_LANGUAGES_COMPILER_ENV_VAR
CMake Error: Error required internal CMake variable not set, cmake may be not be built correctly.
Missing variable is:
CMAKE_LANGUAGES_COMPILER
CMake Error: Could not find cmake module file: /mnt/c/Users/michael/Documents/repos/sweep-sdk/libsweep/build/CMakeFiles/2.8.12.2/CMakeLANGUAGESCompiler.cmake
CMake Error: Could not find cmake module file: CMakeLANGUAGESInformation.cmake
CMake Error: CMAKE_LANGUAGES_COMPILER not set, after EnableLanguage

on WSL ( a close Ubuntu emulation).

`project(sweep C CXX)`

works however

MikeGitb:
I just realized, that the cmake file in examples requires version 3.1, whereas the library cmake file only requires only 2.8.11. Is there a particular reason for that?

MikeGitb:
Also, for the future I think it would be nice if the examples could be build without installing libsweep, but this should not block this PR.

libsweep/CMakeLists.txt
daniel-j-h: Thanks, good catch. I think the explicit `LANGUAGES` is not available in CMake 2.8
https://cmake.org/cmake/help/v2.8.12/cmake.html#command:project

daniel-j-h:
No there was no reason for using CMake 3.1 in the examples. Just lower the requirements there, too. Thanks for catching it!

daniel-j-h:
There are two options for the examples:
- Build them from within the project: which means integrate them into the root `CMakeLists.txt` and let the examples depend and link in `libsweep.so` - maybe even build them by default. This would allow users to build examples without having to install the library. Convenient for sure. The downside here is the examples directory no longer representing what users have to do for integrating `libsweep.so` into their own project.
- Build the examples externally as a free-standing project. We do this currently. Mostly because users can see the `examples/CMakeLists.txt` as a starting point for how their project's `CMakeLists.txt` has to look like. Users can literally copy and paste the examples and take it further from there.

I have no strong opinions here.

daniel-j-h:
@dcyoung can you test on Visual Studio and MinGW - once we confirmed it works there, too, I'd like to merge this into master. Then you can rebase your Windows branch and remove the project files completely.

MikeGitb:
How could it? You have a hardcoded error for windows.

daniel-j-h:
@MikeGitb of course with setting the appropriate compiler and linker flags. It's more about the full build system here.

dcyoung:
With a few tweaks I was able to get the library building on Windows. There is a bit more I need to tinker with. The sweep config can't find the sweep package after installation. And I'm still trying to make sense of the produced visual studio solution. The includes dont yet show up in the solution and I am unfamiliar with the produced project structure.  I need to check out a good  example to see what kind of VS solution structure is supposed to result from cmake. But I'm sure these are all little things that I can fix shortly. I'll be in the office later this afternoon and will try to get it fleshed out.

MikeGitb:
@dcyoung: The project structure that is generated by CMake for VS is indeed not very intuitive but that is (unfortunately) normal. VS2017 has more or less native CMake support and makes user experience much better.
To get the Headerfiles displayed in the project, you have to add them to the `add_library` command (in this case to the `libsweep_SOURCES` vraiable).

daniel-j-h:
I added the public headers to the `add_library` command; they should now appear in your IDE. Please check and adapt accordingly if they still don't show up.

> The sweep config can't find the sweep package after installation.

@dcyoung What do you mean here? The `FindPackage(Sweep)` call in the example's `CMakeLists.txt`? This could be related to the directory where we're installing the `SweepConfig.cmake` to.

dcyoung:
For the missing files in visual studio, adding the other include files (serial.h and protocol.h) by GLOB pattern to the "libsweep_HEADERS" variable did the trick...
`file(GLOB libsweep_HEADERS include/*.h include/sweep/*.h include/sweep/*.hpp)`

@MikeGitb  Should these have been added to libsweep_SOURCES instead?

@daniel-j-h It seems the error is in `find_package_handle_standard_args`. On windows, cmake is installing the sweep library by creating the following directory structure at `C:\Program Files (x86)\sweep\`:

```
├── C:\Program Files (x86)\sweep/
│   ├── include/
│   │   ├── 00_Getting_Started.md
│   │   ├── sweep/
│   │   │   ├── `config.h`
│   │   │   ├── `sweep.h`
│   │   │   ├── `sweep.hpp`
│   ├── lib/
│   │   ├── cmake/
│   │   │   ├── sweep/
│   │   │   │   ├── `SweepConfig.cmake`
│   │   ├── `sweep.dll`
│   │   ├── `sweep.lib`
```

When building the examples, `C:\Program Files (x86)\sweep\lib\cmake\sweep\SweepConfig.cmake` is being executed and errors:

```
CMake Error at C:/Program Files/CMake/share/cmake-3.7/Modules/FindPackageHandleStandardArgs.cmake:138 (message):
  
    Could NOT find Sweep (missing: LIBSWEEP_LIBRARY LIBSWEEP_INCLUDE_DIR)

Call Stack (most recent call first):
    C:/Program Files/CMake/share/cmake-3.7/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  
    C:/Program Files (x86)/sweep/lib/cmake/sweep/SweepConfig.cmake:16 (find_package_handle_standard_args)
  
    CMakeLists.txt:19 (find_package)
```

daniel-j-h:
Hm can you play with the installation path for `SweepConfig.cmake` [here](https://github.com/scanse/sweep-sdk/pull/52/files#diff-35dddea53222a8242e2213c7ef71130aR70).
It looks like we have to figure out a path that is consistent across platforms.

See the documentation on CMake's `FindPackage` and its search paths [here](https://cmake.org/cmake/help/v3.0/command/find_package.html).

This should really work for all three platforms: Linux, Windows, macOS.

MikeGitb:
> Should these have been added to libsweep_SOURCES instead?

@dcyoung: Doesn't matter, as long as they are being added to the library target both are just local variables.

dcyoung:
Hmmm, so the issue was the variables `LIBSWEEP_LIBRARY_DIR` and `LIBSWEEP_INCLUDE_DIR` used by the `SweepConfig.cmake` script were never not initialized. In `examples/CMakeLists.txt` I hard initialized them before the call to `find_package` like so:

```
set(LIBSWEEP_LIBRARY_DIR "C:\\Program Files (x86)\\sweep\\lib\\")
set(LIBSWEEP_INCLUDE_DIR "C:\\Program Files (x86)\\sweep\\include\\")
...
find_package(Sweep REQUIRED)
```
The library (sweep.lib) was found successfully and the examples build just fine. However, running the examples is causing errors because the computer can't find sweep.dll at runtime. Manually moving sweep.dll from the installation directory to the executable directory resolves the issue, and the examples run just fine... but this definitely isn't the way we want to do things. I'm not sure how to pass along the reference to the .dll using cmake. In the original VS integration there was not installation, but instead I was referencing the output of the libsweep library build in each example's project settings. But now we are auto-generating the VS files for the library and the examples separately. Additionally we are supporting building outside of VS, just using cmake and MSVC compiler. 

I am all for this, I just need to figure out how to note the installed .dll location for the examples. 

@MikeGitb Any experience with this?

@daniel-j-h Are you suggesting that we install to a different location? Where in the code are `LIBSWEEP_LIBRARY_DIR` and `LIBSWEEP_INCLUDE_DIR` initialized when running on linux? ... I'll continue looking solution, but any idea how can we avoid this hard code on windows?

daniel-j-h:
> Hmmm, so the issue was the variables LIBSWEEP_LIBRARY_DIR and LIBSWEEP_INCLUDE_DIR used by the SweepConfig.cmake script were never initialized. In examples/CMakeLists.txt I hard initialized them before the call to find_package like so

The idea behind the `find_package(Sweep REQUIRED)` CMake call is exactly that: figure out the library and include dirs and then set these variables for the user. We just have to figure out where to install `SweepConfig.cmake` to on Windows for CMake to pick it up (see link to docs above) - you should never have to set them manually.

Also users can influence variables from the outside by doing

    cmake .. -DKey="Value"

In case installation directories differ.

dcyoung:
Ahhh, that makes a lot more sense. Sorry still getting the hang of CMake here, I appreciate the explanation.

I managed to get things working by adding `C:\Program Files (x86)\sweep\include\sweep\` to the `PATH`. I didn't do this at first, as I was expecting the appropriate `PATH` addition would be `C:\Program Files (x86)\sweep\include` given `SweepConfig.cmake` searches for `sweep/sweep.h` and `sweep/sweep.hpp` instead of just `sweep.h` and `sweep.hpp`. But adding `C:\Program Files (x86)\sweep\include` did NOT work, while adding `C:\Program Files (x86)\sweep\include\sweep` did. 

I'm assuming this is expected behavior and that my expectation was invalid. We'll just have to document this in the README i suppose.

dcyoung:
@daniel-j-h  go ahead and merge. I'll add the windows bits after rebasing the win branch.

PR #57, opened 2017-02-16
Check for scanning device before sending cmd

### LINKED ISSUE #56: Check for !device.is_scanning before sending commands
It looks like the sample rate commit broke the c++ example by placing the calls to get_motor_speed and get_sample_rate AFTER `start_scanning`...

https://github.com/scanse/sweep-sdk/blob/4c12c07c16c6fc82f49fed78929119f1db1e1b07/libsweep/examples/example.cc#L11

This probably works with the dummy library because it sends back information regardless of whether or not its scanning. However, the actual sweep device cannot receive or respond to commands other than `DX\n` while it is scanning. The calls to `get_motor_speed` and `get_sample_rate` need to be placed before `start_scanning`. 

Additionally, in those functions we should probably check if the device is scanning and simply error or report it somehow before trying to message a device... which will only cause more errors. We can do this easily, as we already set a `is_scanning` bool in the `sweep_device` struct

dcyoung:
This PR resolves #56

### DIFF
--- a/libsweep/src/dummy.cc
+++ b/libsweep/src/dummy.cc
@@ -11,7 +11,7 @@ typedef struct sweep_error {
 } sweep_error;
 
 typedef struct sweep_device {
-  bool scanning;
+  bool is_scanning;
   int32_t motor_speed;
   int32_t sample_rate;
   int32_t nth_scan_request;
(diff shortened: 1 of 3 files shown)


daniel-j-h:
Thanks! I think get_scan and reset in both dummy and real impl. are missing checks for scanning and not scanning, respectively.

daniel-j-h:
Fixed and merged! Thanks!

PR #58, opened 2017-02-18
Adds sweep-ctl, resolves #49

### LINKED ISSUE #49: sweep-ctl: interacting with the device from the command line
We really should have some sort of `sweep-ctl` binary to get set and reset the device state.

With an interface similar to

```
$ sweep-ctl /dev/ttyUSB0 get motor_speed
3
$ sweep-ctl /dev/ttyUSB0 set motor_speed 5
5
$ sweep-ctl /dev/ttyUSB0 set motor_speed
3
```

Otherwise users have to modify the examples and build their own binaries for these kind of interactions.

Should get installed in combination with a man page (generated by pandoc).

daniel-j-h:
Adds a `sweep-ctl` binary as described in #49.

Not sure about the man page install dir on macOS and Windows. At the moment I'm using CMake's `GNUInstallDirs` module. Would be great if we can check this first before merging.

### DIFF
--- a/libsweep/CMakeLists.txt
+++ b/libsweep/CMakeLists.txt
@@ -75,6 +75,17 @@ install(DIRECTORY include/sweep DESTINATION include)
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/include/sweep/config.h DESTINATION include/sweep)
 
 
+# sweep-ctl target.
+
+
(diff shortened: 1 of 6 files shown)


MikeGitb:
I don't think there is anything comparable to man pages on windows (except for PowerShell scripts). I'd just put the relevant information into the usage message.

PR #59, opened 2017-02-18
Refactor internal error handling to use exceptions

MikeGitb:
Extracted the exception based error handling from #48 and put it in this separate PR.

Disclaimer: I didn't run any tests on this other than that it compiles (I don't have a hardware device)

### DIFF
--- a/libsweep/include/Error.hpp
+++ b/libsweep/include/Error.hpp
@@ -1,0 +1,7 @@
+#include <stdexcept>
+
+namespace sweep {
+    struct ErrorBase : std::runtime_error {
+        using std::runtime_error::runtime_error;
+    };
+}
--- a/libsweep/include/Error.hpp
+++ b/libsweep/include/Error.hpp
@@ -1,3 +1,6 @@
+#ifndef SWEEP_ERROR_2EADE195E243_H
+#define SWEEP_ERROR_2EADE195E243_H
+
 #include <stdexcept>
 
 namespace sweep {
@@ -5,3 +8,5 @@ namespace sweep {
         using std::runtime_error::runtime_error;
     };
 }
+
+#endif
--- a/libsweep/src/protocol.cc
+++ b/libsweep/src/protocol.cc
@@ -2,13 +2,13 @@
 #include <stdlib.h>
 
 #include "protocol.h"
-#include "../include/Error.hpp"
+#include "error.hpp"
 
 namespace sweep {
 namespace protocol {
 
-struct Error : ::sweep::ErrorBase {
-  using ::sweep::ErrorBase::ErrorBase;
+struct error : ::sweep::error_base {
+  using ::sweep::error_base::error_base;
 };
 
 const uint8_t DATA_ACQUISITION_START[2] = {'D', 'S'};
@@ -83,13 +83,13 @@ void read_response_header(serial::device_s serial, const uint8_t cmd[2], respons
   uint8_t checksum = checksum_response_header(header);
 
   if (checksum != header->cmdSum) {
-    throw Error{"invalid response header checksum"};
+    throw error{"invalid response header checksum"};
   }
 
   bool ok = header->cmdByte1 == cmd[0] && header->cmdByte2 == cmd[1];
 
   if (!ok) {
-    throw Error{"invalid header response commands"};
+    throw error{"invalid header response commands"};
   }
 }
 
@@ -103,13 +103,13 @@ void read_response_param(serial::device_s serial, const uint8_t cmd[2], response
   uint8_t checksum = checksum_response_param(param);
 
   if (checksum != param->cmdSum) {
-    throw Error{"invalid response param header checksum"};
+    throw error{"invalid response param header checksum"};
   }
 
   bool ok = param->cmdByte1 == cmd[0] && param->cmdByte2 == cmd[1];
 
   if (!ok) {
-    throw Error{"invalid param response commands"};
+    throw error{"invalid param response commands"};
   }
 }
 
@@ -122,7 +122,7 @@ void read_response_scan(serial::device_s serial, response_scan_packet_s* scan) {
   uint8_t checksum = checksum_response_scan_packet(scan);
 
   if (checksum != scan->checksum) {
-    throw Error{"invalid scan response commands"};
+    throw error{"invalid scan response commands"};
   }
 }
 
@@ -136,7 +136,7 @@ void read_response_info_motor(serial::device_s serial, const uint8_t cmd[2], res
   bool ok = info->cmdByte1 == cmd[0] && info->cmdByte2 == cmd[1];
 
   if (!ok) {
-    throw Error{"invalid motor info response commands"};
+    throw error{"invalid motor info response commands"};
   }
 }
 
@@ -150,7 +150,7 @@ void read_response_info_sample_rate(sweep::serial::device_s serial, const uint8_
   bool ok = info->cmdByte1 == cmd[0] && info->cmdByte2 == cmd[1];
 
   if (!ok) {
-    throw Error{"invalid sample rate info response commands"};
+    throw error{"invalid sample rate info response commands"};
   }
 }
 
--- a/libsweep/src/protocol.cc
+++ b/libsweep/src/protocol.cc
@@ -2,9 +2,14 @@
 #include <stdlib.h>
 
 #include "protocol.h"
+#include "../include/Error.hpp"
 
 namespace sweep {
 namespace protocol {
+
+struct Error : ::sweep::ErrorBase {
+  using ::sweep::ErrorBase::ErrorBase;
+};
 
 const uint8_t DATA_ACQUISITION_START[2] = {'D', 'S'};
 const uint8_t DATA_ACQUISITION_STOP[2] = {'D', 'X'};
@@ -15,30 +20,6 @@ const uint8_t SAMPLE_RATE_INFORMATION[2] = {'L', 'I'};
 const uint8_t VERSION_INFORMATION[2] = {'I', 'V'};
 const uint8_t DEVICE_INFORMATION[2] = {'I', 'D'};
 const uint8_t RESET_DEVICE[2] = {'R', 'R'};
-
-typedef struct error {
-  const char* what; // always literal, do not deallocate
-} error;
-
-// Constructor hidden from users
-static error_s error_construct(const char* what) {
-  SWEEP_ASSERT(what);
-
-  auto out = new error{what};
-  return out;
-}
-
-const char* error_message(error_s error) {
-  SWEEP_ASSERT(error);
-
-  return error->what;
-}
-
-void error_destruct(error_s error) {
-  SWEEP_ASSERT(error);
-
-  delete error;
-}
 
 static uint8_t checksum_response_header(response_header_s* v) {
   SWEEP_ASSERT(v);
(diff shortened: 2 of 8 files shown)


libsweep/include/Error.hpp
```diff
@@ -0,0 +1,12 @@
+#ifndef SWEEP_ERROR_2EADE195E243_H
+#define SWEEP_ERROR_2EADE195E243_H
+
+#include <stdexcept>
+
+namespace sweep {
+    struct ErrorBase : std::runtime_error {
```
daniel-j-h: Can you please lowercase the file to `error.h` as well as `error` here: it's already in the `sweep` namespace, so it's the generic `sweep::error` type.

libsweep/src/protocol.cc
```diff
@@ -2,10 +2,15 @@
 #include <stdlib.h>
 
 #include "protocol.h"
+#include "../include/Error.hpp"
 
 namespace sweep {
 namespace protocol {
 
+struct Error : ::sweep::ErrorBase {
```
daniel-j-h: Same here

libsweep/src/protocol.cc
```diff
@@ -2,10 +2,15 @@
 #include <stdlib.h>
 
 #include "protocol.h"
+#include "../include/Error.hpp"
```
daniel-j-h: Please no relative include paths, the include dir should be on the search path already (check CMakeLists.txt)

libsweep/src/sweep.cc
```diff
@@ -1,15 +1,18 @@
 #include "sweep.h"
 #include "protocol.h"
 #include "serial.h"
+#include "../include/Error.hpp"
```
daniel-j-h: Same here

libsweep/src/sweep.cc
```diff
@@ -19,25 +22,17 @@ typedef struct sweep_device {
 
 #define SWEEP_MAX_SAMPLES 4096
 
-typedef struct sweep_scan {
+struct sweep_scan {
```
daniel-j-h: Hm if you want to remove typedefs feel free to do it for all structs

libsweep/src/sweep.cc
```diff
@@ -46,149 +41,102 @@ void sweep_error_destruct(sweep_error_s error) {
...
-
-sweep_device_s sweep_device_construct(const char* port, int32_t bitrate, sweep_error_s* error) {
+// internal implementations of functions - can throw exceptions
+namespace {
+
+struct Error : ::sweep::ErrorBase {
+  using ::sweep::ErrorBase::ErrorBase;
+};
+
+// forward declarations for internal definitions
```
daniel-j-h: Why? We include the `sweep.h` header anyway, no?

libsweep/src/sweep.cc
```diff
@@ -46,149 +41,102 @@ void sweep_error_destruct(sweep_error_s error) {
...
-  return sweep_device_construct("/dev/ttyUSB0", 115200, error);
-}
-
-sweep_device_s sweep_device_construct(const char* port, int32_t bitrate, sweep_error_s* error) {
+// internal implementations of functions - can throw exceptions
+namespace {
+
+struct Error : ::sweep::ErrorBase {
+  using ::sweep::ErrorBase::ErrorBase;
+};
```
daniel-j-h: Shouldn't we use the base from error.h here?

libsweep/src/sweep.cc
```diff
@@ -46,149 +41,102 @@ void sweep_error_destruct(sweep_error_s error) {
...
-    sweep::serial::error_destruct(serialerror);
-    return nullptr;
-  }
 
-  auto out = new sweep_device{serial, /*is_scanning=*/true};
+  sweep::serial::device_s serial = sweep::serial::device_construct(port, bitrate);
 
-  sweep_error_s stoperror = nullptr;
-  sweep_device_stop_scanning(out, &stoperror);
+  auto out = std::unique_ptr<sweep_device>(new sweep_device{serial, /*is_scanning=*/true});
```
daniel-j-h: Why the unique ptr here? For cleaning up in case of exceptions? This is a bit ugly

libsweep/src/sweep.cc
```diff
@@ -46,149 +41,102 @@ void sweep_error_destruct(sweep_error_s error) {
...
-    return nullptr;
-  }
+  sweep_device_stop_scanning(out.get());
 
-  return out;
+  return out.release();
 }
 
-void sweep_device_destruct(sweep_device_s device) {
+void sweep_device_destruct_impl(sweep_device_s device) {
```
daniel-j-h: Why the change to _impl here?

libsweep/src/sweep.cc
```diff
@@ -212,7 +160,7 @@ sweep_scan_s sweep_device_get_scan(sweep_device_s device, sweep_error_s* error)
     }
   }
 
-  auto out = new sweep_scan;
+  auto out = std::unique_ptr<sweep_scan>(new sweep_scan);
```
daniel-j-h: Same here

libsweep/src/sweep.cc
```diff
@@ -395,38 +262,132 @@ void sweep_device_set_sample_rate(sweep_device_s device, int32_t hz, sweep_error
...
-  sweep::protocol::response_param_s response;
-  sweep::protocol::read_response_param(device->serial, sweep::protocol::SAMPLE_RATE_ADJUST, &response, &protocolerror);
+  sweep::protocol::write_command(device->serial, sweep::protocol::RESET_DEVICE);
+}
 
-  if (protocolerror) {
-    *error = sweep_error_construct("unable to receive sample rate command response");
-    sweep::protocol::error_destruct(protocolerror);
-    return;
+template <class F> auto translateException(F&& f, sweep_error_s* error) -> decltype(std::forward<F>(f)()) {
```
daniel-j-h: The following should probably be in the error.h header

libsweep/src/sweep.cc
```diff
@@ -395,38 +262,132 @@ void sweep_device_set_sample_rate(sweep_device_s device, int32_t hz, sweep_error
...
+  try {
+    return std::forward<F>(f)();
+  } catch (const sweep::ErrorBase& e) {
+    *error = new sweep_error{e.what()};
+  } catch (const std::bad_alloc&) {
```
daniel-j-h: Why not catching `std::exception&` for which there is a `.what()`

libsweep/src/sweep.cc
```diff
@@ -395,38 +262,132 @@ void sweep_device_set_sample_rate(sweep_device_s device, int32_t hz, sweep_error
...
+    *error = new sweep_error{"Could not allocate enough memory"};
+  } catch (...) {
+    *error = new sweep_error{"Unknown exception"};
+  }
+  return {};
+}
+
+template <class F> void translateExceptionV(F&& f, sweep_error_s* error) {
```
daniel-j-h: Would be great if we can overload this function instead

libsweep/src/sweep.cc
```diff
@@ -395,38 +262,132 @@ void sweep_device_set_sample_rate(sweep_device_s device, int32_t hz, sweep_error
...
+    *error = new sweep_error{e.what()};
+  } catch (const std::bad_alloc&) {
+    *error = new sweep_error{"Could not allocate enough memory"};
+  } catch (...) {
+    *error = new sweep_error{"Unknown exception"};
   }
 }
+} // local namespace
+
+/* ###################### implementation of public API ################################# */
```
daniel-j-h: Ah now I understand - you split off the C API and the internal functions. Hm I see no benefit in doing so, what's the reason behind this?

libsweep/src/unix/serial.cc
```diff
@@ -1,6 +1,7 @@
 #define _POSIX_C_SOURCE 200809L
 
 #include "serial.h"
+#include "../../include/Error.hpp"
```
daniel-j-h: Same here

libsweep/include/Error.hpp
MikeGitb: OK on the lower case, but removing `Base` will let it collide with the error defined in sweep.cc.

libsweep/src/sweep.cc
MikeGitb: Separation of concerns: The private functions are concerned with implementing the functionality the implementation of the public functions then only needs to take care about eror translation and any other adaptions to the ABI that might be necessary in the future. Also `translateException` needs some function / function object to be passed to it anyway, so the only alternative I see would be to wrapp the internal functionality into a lambda which doesn't really make the code more readable.

libsweep/src/sweep.cc
MikeGitb: As I said in the other thread: If you see a simple way to do this be my guest. also I really don't see what the advantage would be.

libsweep/src/sweep.cc
MikeGitb: Overlooked, will fix

libsweep/src/sweep.cc
MikeGitb: As it is translating an exception into and ABI error, which is defined in `sweep.h` and no other part of the code has to care about this I don't see why this should be moved there.

libsweep/src/sweep.cc
MikeGitb: To avoid naming collision (the API I version doesn't hav an error parameter to distinguish between the two). See comment below about why I have two in the first place

libsweep/src/sweep.cc
MikeGitb: Precisely. `sweep_device_stop_scanning` can throw and besides it is my policy to not have a naked `new` anywhere in my code if I can at all help it. 

What would be your suggestion (in c++14 you can use make_unique)?

libsweep/src/sweep.cc
MikeGitb: See comment at the end

libsweep/src/sweep.cc
MikeGitb: That would also be a possibility. 
In general I wanted to distinguish between exceptions thrown in the library in general and exceptions that are thrown on this particular level (e.g. as opposed to the IO or protocol level). But currently there is admittedly nothing that makes use of that distinction. We could even use the same exception type everywhere in the code.

libsweep/src/sweep.cc
MikeGitb: Must have slipped through, will fix. 
I originally had all the typedefs removed throughout the library but wanted to keep this PR clean of it (there is already going on quite a lot anyway), so I reverted those changes - except that I apparently overlooked this one.

libsweep/src/protocol.cc
MikeGitb: Can do.
Although I'm really no fan of unqualified includes for files outside the same directory - in particular for files with very generic names.

MikeGitb:
@daniel-j-h: Ok, I
- made `error.hpp` and `error` lower case
- changed to unqualified includes
- changed the error translation a bit

I did not incorporate the rest of your feedback and am waiting for your responses to my comments instead. I'm not sure, why git marks them all as outdated

libsweep/src/sweep.cc
MikeGitb: Actually I think I'll just leave it for now and provide another PR once / if this one gets merged

MikeGitb:
Rebased onto master

MikeGitb:
Rebased on master and rewrote error handling in win/serial.cc to use exceptions

MikeGitb:
So ist there any chance this will get merged in the near future?

daniel-j-h:
Yes, please excuse my inactivity I'm out until late next week. Maybe
@dcyoung can jump in here - otherwise you will have to wait a bit. I
want to get us to switch to exceptions and references for sure.

MikeGitb:
No rush. Main question is, if you are OK with the general structure of this PR as it is now.

MikeGitb:
Closing this for now - I think after the last round of changes a merge is no longer possible. I won't delete the branch though, as parts might be salvageable for future PRs

Issue #16, opened 2016-11-12
Improve get_scan throughput


daniel-j-h:
At the moment `get_scan` waits for the sync bit then records responses until the next sync bit. This makes sure we gather and report only full 360 degree scans:

https://github.com/scanse/sweep-sdk/blob/853f113543621daf6ce8553177f7d2153d2657c4/libsweep/sweep.c#L177-L231

But it's also wasteful since we're throwing away the data before the first sync bit, limiting our throughput.

To keep the low-level library simple I don't want to go the route of a worker thread, accumulating behind the scenes. Instead, we should be able to check for the angle we start recording at and stop recording as soon as we reach it again.

We don't guarantee for the first sample to start at angle 0 degree after all. If the user needs this, she can accumulate herself.

daniel-j-h:
Looks like the Python bindings fail to keep up with polling in some use-cases.

I could see read errors when using threads in Python for polling + clustering + drawing.

This means, we have to put back the threaded worker into `libsweep` for accumulating scan responses.

dcyoung:
I'm not sure what you mean by "throughput" here. Could you clarify? Trashing all the data before the first error/sync byte will mean that data collected in-between the receipt header and the first sync byte will go unreported. Functionally it would present to the host as a slight delay before the sensor starts to acquire data. But once the first sync byte is noted, the throughput is consistent yes? 

The number of sensor readings in a scan is variable, so any user application will have to support variable number of data points in a scan. As long as we package them in the same format as complete scans, is there a problem with reporting partial scans? Thinking ahead, if we add a very slow motor speed (<<1Hz), it could be unwise to limit reporting to only complete scans.  I could see some applications wanting a guarantee of a full scan however, so perhaps we should consider a setting that toggles the reporting of partial scans. Ie: if the user sends a "DS" command with the setting enabled, then the first partial scan is included, whereas the partial scan is trashed if the setting is disabled. 

Additionally, for efficiency, we may want to modify `SWEEP_MAX_SAMPLES` based on the device's current settings, given that various combinations of motor speed and sample rate will produce drastically different point counts in a single scan.

daniel-j-h:
See [here](https://github.com/scanse/sweep-sdk/blob/853f113543621daf6ce8553177f7d2153d2657c4/libsweep/sweep.c#L177-L231). At the moment we
- wait for the sync bit
- start recording
- wait for another sync bit to indicate the end   <- (1)
- return this as a full scan
- repeat

[...]

PR #60, opened 2017-02-19
Improve throughput

MikeGitb:
This solves the problem of `sweep_device_get_scan` skiping every second scan mentioned in #16. 

Every call to `sweep_device_get_scan` now simply returns all packets up to and including the next packet that has the sync bit set. When the function is called in a loop this should always return a full 360° scan except for the first call and it is up to the user to decide, whether or not he wants to discard that first scan. This is also pretty much the semantics needed when using a worker thread.

As a sideffect it also simplifies the code a bit ;)

### DIFF
--- a/libsweep/src/sweep.cc
+++ b/libsweep/src/sweep.cc
@@ -178,10 +178,9 @@ sweep_scan_s sweep_device_get_scan(sweep_device_s device, sweep_error_s* error) 
 
   sweep::protocol::response_scan_packet_s responses[SWEEP_MAX_SAMPLES];
 
-  int32_t first = 0;
-  int32_t last = SWEEP_MAX_SAMPLES;
+  int32_t received = 0;
 
-  for (int32_t received = 0; received < SWEEP_MAX_SAMPLES; ++received) {
+  while (received < SWEEP_MAX_SAMPLES) {
     sweep::protocol::read_response_scan(device->serial, &responses[received], &protocolerror);
 
     if (protocolerror) {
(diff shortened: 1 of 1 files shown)


daniel-j-h:
Thank you! Just tested with device - works as advertised :)

dcyoung:
I haven't tried this out yet, but from the description it sounds like the last entry of the returned scan is the sync reading of the following scan. If this is the case, we really should be discarding and not including this last entry as it will disturb anything that assumes readings of a scan are arranged in order of increasing degree. Any attempt to order the readings by azimuth would then result in readings that are out of chronological order. Until we get the worker thread and queue implementation in place, we should not include the sync reading.

MikeGitb:
@dcyoung: 
> I haven't tried this out yet, but from the description it sounds like the last entry of the returned scan is the sync reading of the following scan.

Correct.

> as it will disturb anything that assumes readings of a scan are arranged in order of increasing degree

Is this something you want to guarantee? If so, instead of dropping the sync reading we can store it for the next call to `get_scan`. Makes the code a bit more complicated, but nothing I'd call troublesome.

dcyoung:
We definitely want to guarantee that a scan will contain only the readings accrued during a single rotation. I think that is more important than avoiding the loss of a single reading. 

Part of the SDK design was to avoid any persistent or global variables. Seems like holding onto a single reading between scans would conflict with that. And it would be rather ugly. I don't see much harm in losing a single reading until the scan worker is in place, which should hopefully be ASAP.

MikeGitb:
@dcyoung: 

> We definitely want to guarantee that a scan will contain only the readings accrued during a single rotation. I think that is more important than avoiding the loss of a single reading.

Are you defining a rotation as starting at 0 degree and ending at 360-epsilon? Or any interval that spans 360degrees? Neither the current implementation nor my proposed fix would return more than 360 degrees. But yes currently one scan is (0, 0] instead of [0, 0).

> Part of the SDK design was to avoid any persistent or global variables. Seems like holding onto a single reading between scans would conflict with that. And it would be rather ugly.

well, it would only require one member variable in sweep_device and the logic would not be more complicated than what it looked before this PR. But if this is a temporal fix anyway I agree - lets just dump it.

> I don't see much harm in losing a single reading until the scan worker is in place, which should hopefully be ASAP.

Implementing that is rather trivial (In fact I have an implementation ready), but someone has to decide on what the interface should look like (simplest case would be to keep the current interface and make threading a pure implementation detail).

dcyoung:
@MikeGitb 
Yes, our definition of rotation is defined as 0deg -> 360 deg, which matches the devices physical 0 degree marker. While every application could ideally parse and process the readings to handle them intelligently, we want to make it as simple as possible for users to interpret the return from `get_scan`.

I agree that the worker thread should be hidden in the implementation.

MikeGitb:
@dcyoung: Well as I said in other threads, I generally prefer to give the User control over if they want to use threads or not (and possibly even allow them to provide the thread themselves). But if simplicity is most important to you I can probably create a PR.

dcyoung:
I'm not against allowing the user to specify the use of threading, but we'd definitely want the base behavior (ie: the simple versions of everything) to use threading in their implementation. This is a bit off topic for this  PR though. Let's keep that discussion in #31

PR #63, opened 2017-03-26
sweeppy: fix Python 3 compatibility

simon04:
Rationale: `ctypes.c_char_p` map C `char*` to Python byte objects.
Thus, encoding Python strings is required before passing it as argument.

This fixes a `opening serial port failed` problem when using Python 3.


daniel-j-h:
Ouch, this is a sneaky one! Thanks for catching it!

PR #70, opened 2017-04-14
Updated Protocol + Scan worker

dcyoung:
This PR brings a few changes:

1. Updated protocol to match the new firmware that is installed on devices shipping out today.
2. Updated to perform checks for status byte codes indicating failures that do not result in errors. 
3. Provide methods (and update existing methods) to wait until the device is ready before attempting to change motor speed or start scanning.
4. Add scan worker to accumulate scans in a queue using a background thread.
5. Return complete scans between 0-360 degrees (other than first scan which is inherently partial).
6. Offer alternative non-blocking functions where possible (currently only available in C library).
7. Increase delay between stop commands in the stop_scanning function to account for the device logging session data after responding to the first stop.

I have tested the examples, and the bindings on windows. I have tested the examples and bindings on ubuntu 14.04, but @kent-williams mentioned he might have experienced an issue with the python bindings on 16.04. The travis builds might not indicate anything as they're using the dummy build. So i'd love to get additional testing feedback if you have the time @daniel-j-h 

This PR needs get merged ASAP, as people will be receiving devices very soon (early/mid next week). I'll go ahead and merge this in a few days if there aren't any complaints. But please add your feedback if you have it!

### DIFF
--- a/libsweep/README.md
+++ b/libsweep/README.md
@@ -185,20 +185,42 @@ void sweep_device_start_scanning(sweep_device_s device, sweep_error_s* error)
 ```
 
 Signals the `sweep_device_s` to start scanning.
+If the motor is stationary (0Hz), will automatically set motor speed to default 5Hz.
+Will block until the the motor speed is stable (uses `sweep_device_wait_until_motor_ready` internally).
+Starts internal background thread to accumulate and queue up scans. Scans can then be retrieved using `sweep_device_get_scan`.
 In case of error a `sweep_error_s` will be written into `error`.
+
 
 ```c++
 void sweep_device_stop_scanning(sweep_device_s device, sweep_error_s* error)
 ```
 
 Signals the `sweep_device_s` to stop scanning.
+Blocks for ~35ms to allow time for the trailing data stream to collect and flush internally, before sending a second stop command and validate the response.
+In case of error a `sweep_error_s` will be written into `error`.
+
+```c++
+void sweep_device_wait_until_motor_ready(sweep_device_s device, sweep_error_s* error)
+```
+
+Blocks until the `sweep_device_s` is ready, or the method times out (after 8 seconds). A device is ready when the motor speed has stabilized to the current setting, and the calibration routine is complete. The worst case wait time is around 6 seconds, which includes both motor stabilization and calibration. For visual reference, the blue LED on the device will blink unil the device is ready. This method is useful when the device is powered on, or when adjusting motor speed. If the device is NOT ready, it will respond to certain commands (`DS` or `MS`) with a status code indicating a failure to execute the command. Therefore, it is best practice to avoid this entirely by calling `sweep_device_wait_until_motor_ready` before calling a command that requires a ready device.
+In case of error a `sweep_error_s` will be written into `error`.
+
+```c++
+bool sweep_device_get_motor_ready(sweep_device_s device, sweep_error_s* error)
+```
+
+Returns `true` if the device is ready. A device is ready if the motor speed has stabilized to the current setting, and the calibration routine is complete. 
+This method can be used to create a non-blocking alternative to `sweep_device_wait_until_motor_ready` in user programs.
+For visual reference, the blue LED on the device will blink during calibration/speed stabilization, and stop blinking when the device is ready.
 In case of error a `sweep_error_s` will be written into `error`.
 
 ```c++
 int32_t sweep_device_get_motor_speed(sweep_device_s device, sweep_error_s* error)
 ```
 
-Returns the `sweep_device_s`'s motor speed in Hz.
+Returns the `sweep_device_s`'s motor speed setting in Hz.
+If the motor speed is currently changing, the returned motor speed is the target speed at which the device will stabilize.
 In case of error a `sweep_error_s` will be written into `error`.
 
 ```c++
@@ -206,8 +228,10 @@ void sweep_device_set_motor_speed(sweep_device_s device, int32_t hz, sweep_error
 ```
 
 Sets the `sweep_device_s`'s motor speed in Hz.
-The device supports speeds of 1 Hz to 10 Hz.
+Blocks until prior motor speed has stabilized.
+The device supports speeds of 0 Hz to 10 Hz, but be careful that the device is not set at 0Hz before calling `sweep_device_start_scanning`.
 In case of error a `sweep_error_s` will be written into `error`.
+
 
 ```c++
 int32_t sweep_device_get_sample_rate(sweep_device_s device, sweep_error_s* error)
@@ -221,8 +245,8 @@ void sweep_device_set_sample_rate(sweep_device_s device, int32_t hz, sweep_error
 ```
 
 Sets the `sweep_device_s`'s sample rate in Hz.
-The device supports sample rates of 500 Hz, 750 Hz and 1000 Hz.
-The device guarantees for those sample rates but they can be slightly higher by a maximum of roughly 50-100 Hz.
+The device supports setting sample rate to the following values: 500 Hz, 750 Hz and 1000 Hz.
+These sample rates are not exact. They are general ballpark values. The actual sample rate may differ slightly.
 In case of error a `sweep_error_s` will be written into `error`.
 
 ```c++
@@ -245,8 +269,10 @@ Opaque type representing a single full 360 degree scan from a `sweep_device_s`.
 sweep_scan_s sweep_device_get_scan(sweep_device_s device, sweep_error_s* error)
 ```
 
-Blocks waiting for the `sweep_device_s` to accumulate a full 360 degree scan into `sweep_scan_s`.
+Returns the ordered readings (1st to last) from a single scan.
+Retrieves the oldest scan from a queue of scans accumulated in a background thread. Blocks until a scan is available. To be used after calling `sweep_device_start_scanning`.
 In case of error a `sweep_error_s` will be written into `error`.
+
 
 ```c++
 void sweep_scan_destruct(sweep_scan_s scan)
@@ -278,7 +304,40 @@ int32_t sweep_scan_get_signal_strength(sweep_scan_s scan, int32_t sample)
 
 Returns the signal strength (0 low -- 255 high) for the `sample`th sample in the `sweep_scan_s`.
 
+<!--
+#### Alternative direct functions
+Many of the SDK methods block for various reasons, such as waiting for a data stream to terminate, waiting for the device to finish calibrating, or waiting for the motor speed to stabilize. This can be an issue for low-level applications where blocking isn't ideal. The following methods (mostly non-blocking) are provided as a workaround. Care must be taken when using these methods as they can error on valid failures if the state of the device is not checked before hand.
 
+```c++
+void sweep_device_attempt_start_scanning(sweep_device_s device, sweep_error_s* error)
+```
+
+Non-blocking alternative to `sweep_device_start_scanning`. 
+Signals the `sweep_device_s` to start scanning. 
+Will only succeed if the motor speed is \>0Hz and stable.
+User is responsible for checking these conditions before calling. User can check that motor speed has stabilized using `sweep_device_get_motor_ready` and that motor speed is \> 0Hz using `sweep_device_get_speed`.
+Will NOT start an internal background thread. User is responsible for keeping up with incoming scans by calling `sweep_device_get_scan_direct`.
+In case of error a `sweep_error_s` will be written into `error`. This method will error on legitimate failures (ex: the motor speed is stationary or has not yet stabilized).
+
+
+```c++
+sweep_scan_s sweep_device_get_scan_direct(sweep_device_s device, sweep_error_s* error)
+```
+
+Returns the ordered readings from the 2nd reading of the current scan through the 1st reading of the subsequent scan.
+Blocks waiting for the `sweep_device_s` to accumulate a full 360 degree scan into `sweep_scan_s`. To be used after calling `sweep_device_attempt_start_scanning`, NOT after `sweep_device_start_scanning`.
+In case of error a `sweep_error_s` will be written into `error`.
+
+```c++
+void sweep_device_attempt_set_motor_speed(sweep_device_s device, int32_t hz, sweep_error_s* error)
+```
+
+Non-blocking alternative to `sweep_device_set_motor_speed`.
+Sets the `sweep_device_s`'s motor speed in Hz.
+Will only succeed if the device is ready. A device is only ready if the motor speed has stabilized to the current setting and the calibration routine is complete.
+Proper use involves checking that the motor speed has stabilized (using `sweep_device_get_motor_ready`) before attempting to adjust the motor speed to a new value.
+In case of error a `sweep_error_s` will be written into `error`. This method will error on legitimate failures (ex: the motor speed has not yet stabilized).
+-->
 ### License
 
 Copyright © 2016 Daniel J. Hofmann
--- a/libsweep/README.md
+++ b/libsweep/README.md
@@ -2,6 +2,8 @@
 
 Low-level Scanse Sweep LiDAR library. Comes as C99 library `sweep.h` with optional C++11 header `sweep.hpp` on top of it.
 
+
+
 ### Quick Start
 
 ```bash
@@ -87,6 +89,12 @@ Table of Contents:
 
 
 #### Version And ABI Management
+Compatibility:
+
+| libsweep | sweep firmware |
+| -------- | :------------: |
+| v1.x.x   | v1.1           |
+| v0.x.x   | v1.0           |
 
 ```c++
 SWEEP_VERSION_MAJOR
--- a/libsweep/include/protocol.h
+++ b/libsweep/include/protocol.h
@@ -24,6 +24,7 @@ void error_destruct(error_s error);
 extern const uint8_t DATA_ACQUISITION_START[2];
 extern const uint8_t DATA_ACQUISITION_STOP[2];
 extern const uint8_t MOTOR_SPEED_ADJUST[2];
+extern const uint8_t MOTOR_READY[2];
 extern const uint8_t MOTOR_INFORMATION[2];
 extern const uint8_t SAMPLE_RATE_ADJUST[2];
 extern const uint8_t SAMPLE_RATE_INFORMATION[2];
@@ -136,11 +137,20 @@ static_assert(sizeof(response_info_version_s) == 21, "response info version size
 typedef struct {
   uint8_t cmdByte1;
   uint8_t cmdByte2;
+  uint8_t motor_ready[2];
+  uint8_t term;
+} response_info_motor_ready_s;
+
+static_assert(sizeof(response_info_motor_ready_s) == 5, "response info motor ready size mismatch");
+
+typedef struct {
+  uint8_t cmdByte1;
+  uint8_t cmdByte2;
   uint8_t motor_speed[2];
   uint8_t term;
-} response_info_motor_s;
+} response_info_motor_speed_s;
 
-static_assert(sizeof(response_info_motor_s) == 5, "response info motor size mismatch");
+static_assert(sizeof(response_info_motor_speed_s) == 5, "response info motor speed size mismatch");
 
 typedef struct {
   uint8_t cmdByte1;
@@ -149,7 +159,7 @@ typedef struct {
   uint8_t term;
 } response_info_sample_rate_s;
 
-static_assert(sizeof(response_info_sample_rate_s) == 5, "response info sample rate siye mismatch");
+static_assert(sizeof(response_info_sample_rate_s) == 5, "response info sample rate size mismatch");
 
 // Done with in-memory representations for packets we send over the wire.
 #pragma pack(pop)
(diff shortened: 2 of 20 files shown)


daniel-j-h:
Thanks for the heads up, I'll be on a plane the next day(s), can jump back into this not earlier than maybe Tuesday or Wednesday.

What's the Ubuntu 16.04 issue with Python? Check the open PRs there should be one with Docker (sorry on a phone). You could use a local 16.04 container and try to reproduce.

libsweep/README.md
```diff
@@ -185,31 +185,75 @@ void sweep_device_start_scanning(sweep_device_s device, sweep_error_s* error)
...
 
 ```c++
+void sweep_device_attempt_start_scanning(sweep_device_s device, sweep_error_s* error)
+```
+
+Non-blocking alternative to `sweep_device_start_scanning`. 
+Signals the `sweep_device_s` to start scanning. 
+Will only succeed if the motor speed is \>0Hz and stable.
+User is responsible for checking these conditions before calling. User can check that motor speed has stabilized using `sweep_device_get_motor_ready` and that motor speed is \> 0Hz using `sweep_device_get_speed`.
+Will NOT start an internal background thread. User is responsible for keeping up with incoming scans by calling `sweep_device_get_scan_direct`.
```
daniel-j-h: Hmm why do we mix waiting for motor to stabilize and worker thread here? (And in the PR in general)
Shouldn't the two be orthogonal concepts?

What I'm a bit worried about is the API is starting to get much more confusing than just start scanning - get scans.

libsweep/README.md
```diff
@@ -185,31 +185,75 @@ void sweep_device_start_scanning(sweep_device_s device, sweep_error_s* error)
...
+In case of error a `sweep_error_s` will be written into `error`. This method will error on legitimate failures (ex: the motor speed is stationary or has not yet stabilized).
+
+```c++
 void sweep_device_stop_scanning(sweep_device_s device, sweep_error_s* error)
 ```
 
 Signals the `sweep_device_s` to stop scanning.
+Blocks for ~20ms to allow time for trailing data stream to collect and flush internally.
```
daniel-j-h: Does this depend on motor speed? 20ms is the upper bound?

libsweep/README.md
```diff
@@ -185,31 +185,75 @@ void sweep_device_start_scanning(sweep_device_s device, sweep_error_s* error)
...
+In case of error a `sweep_error_s` will be written into `error`.
+
+```c++
+void sweep_device_wait_until_motor_ready(sweep_device_s device, sweep_error_s* error)
+```
+
+Blocks until the `sweep_device_s`'s motor speed has stabilized. This is useful whenever the motor speed is changed. The worst case is a 6 second wait for motor stabilization and calibration. If the motor is already spinning at the desired speed, this returns almost instantly. For visual reference, the blue LED on the device will blink while the motor speed is changing, and stop blinking when the speed has stabilized. If the motor speed has NOT yet stabilized, the device will respond to certain commands (`DS` or `MS`) with a status code indicating a failure to execute the command. Therefore, it is best practice to avoid this entirely by calling `sweep_device_wait_until_motor_ready` before calling a command that requires motor speed is stabilized.
+In case of error a `sweep_error_s` will be written into `error`. Will timeout.
```
daniel-j-h: What does "will timeout" mean here? That there is an internal timeout of 6s and it will error out if this limit is exceeded?

libsweep/README.md
```diff
@@ -185,31 +185,75 @@ void sweep_device_start_scanning(sweep_device_s device, sweep_error_s* error)
...
+
+```c++
+int32_t sweep_device_get_motor_ready(sweep_device_s device, sweep_error_s* error)
+```
+
+Returns 0 if `sweep_device_s`'s motor speed has stabilized. 
+Returns 1 if not.
```
daniel-j-h: We require C99+ in the API, so we have `stdbool.h` - please use bool here.

https://github.com/scanse/sweep-sdk/blob/master/libsweep/include/sweep/sweep.h#L42

`bool sweep_device_is_motor_ready(..)`

libsweep/README.md
```diff
@@ -245,7 +289,16 @@ Opaque type representing a single full 360 degree scan from a `sweep_device_s`.
 sweep_scan_s sweep_device_get_scan(sweep_device_s device, sweep_error_s* error)
 ```
 
-Blocks waiting for the `sweep_device_s` to accumulate a full 360 degree scan into `sweep_scan_s`.
+Returns the ordered readings (1st to last) from a single scan.
+Retrieves the oldest scan from a queue of scans accumulated in a background thread. Blocks until a scan is available. To be used after calling `sweep_device_start_scanning`, NOT after `sweep_device_attempt_start_scanning`.
```
daniel-j-h: This and below is what I meant - there now is a direct dependency between functions. And the naming

    get_scan vs get_scan_direct
    start_scan vs attempt_start_scan

does not really make it clear when to use which version.

daniel-j-h: Do we give any guarantees wrt. the queue? How large is it? Should the user be able to specify?

libsweep/include/protocol.h
```diff
@@ -136,11 +137,20 @@ static_assert(sizeof(response_info_version_s) == 21, "response info version size
 typedef struct {
   uint8_t cmdByte1;
   uint8_t cmdByte2;
+  uint8_t motor_ready[2];
+  uint8_t term;
+} response_info_motor_ready_s;
+
+static_assert(sizeof(response_info_motor_ready_s) == 5, "response info motor ready size mismatch");
```
daniel-j-h: This means the SDK will no longer work on old devices, no? If so we need to increment the lib's major version.

libsweep/include/protocol.h
```diff
@@ -166,7 +176,11 @@ void read_response_param(sweep::serial::device_s serial, const uint8_t cmd[2], r
 
 void read_response_scan(sweep::serial::device_s serial, response_scan_packet_s* scan, error_s* error);
 
-void read_response_info_motor(sweep::serial::device_s serial, const uint8_t cmd[2], response_info_motor_s* info, error_s* error);
+void read_response_info_motor_ready(sweep::serial::device_s serial, const uint8_t cmd[2], response_info_motor_ready_s* info,
+                                    error_s* error);
+
+void read_response_info_motor_speed(sweep::serial::device_s serial, const uint8_t cmd[2], response_info_motor_speed_s* info,
+                                    error_s* error);
```
daniel-j-h: Argh sorry this is still in the out-error style; I haven't had the time yet to integrate @MikeGitb's refactoring to use exceptions, sorry.

libsweep/include/sweep/sweep.h
```diff
@@ -48,27 +48,43 @@ typedef struct sweep_scan* sweep_scan_s;
...
-SWEEP_API sweep_device_s sweep_device_construct_simple(const char* port, sweep_error_s* error);
-SWEEP_API sweep_device_s sweep_device_construct(const char* port, int32_t bitrate, sweep_error_s* error);
-SWEEP_API void sweep_device_destruct(sweep_device_s device);
+// ------------------ Blocking Methods (blocking) ---------------- //
 
 SWEEP_API void sweep_device_start_scanning(sweep_device_s device, sweep_error_s* error);
 SWEEP_API void sweep_device_stop_scanning(sweep_device_s device, sweep_error_s* error);
 
+SWEEP_API void sweep_device_wait_until_motor_ready(sweep_device_s device, sweep_error_s* error);
+SWEEP_API void sweep_device_set_motor_speed(sweep_device_s device, int32_t hz, sweep_error_s* error);
```
daniel-j-h: Should set_motor_speed block until motor is ready?

libsweep/include/sweep/sweep.h
```diff
@@ -48,27 +48,43 @@ typedef struct sweep_scan* sweep_scan_s;
...
+
+// Gathers a scan directly (not retrieving from the queue)
+SWEEP_API sweep_scan_s sweep_device_get_scan_direct(sweep_device_s device, sweep_error_s* error);
+// Retrieves a scan from the queue (will block until scan is available)
 SWEEP_API sweep_scan_s sweep_device_get_scan(sweep_device_s device, sweep_error_s* error);
-SWEEP_API void sweep_scan_destruct(sweep_scan_s scan);
+// Accumulates scans in the queue  (method to be used by background thread)
+SWEEP_API void sweep_device_accumulate_scans(sweep_device_s device);
```
daniel-j-h: Why do we expose this to the user? How is the user supposed to call this?

libsweep/include/sweep/sweep.hpp
```diff
@@ -87,15 +89,24 @@ sweep::sweep(const char* port, std::int32_t bitrate)
...
+
 std::int32_t sweep::get_motor_speed() { return ::sweep_device_get_motor_speed(device.get(), detail::error_to_exception{}); }
 
 void sweep::set_motor_speed(std::int32_t speed) {
   ::sweep_device_set_motor_speed(device.get(), speed, detail::error_to_exception{});
 }
 
+void sweep::attempt_set_motor_speed(std::int32_t speed) {
+  ::sweep_device_attempt_set_motor_speed(device.get(), speed, detail::error_to_exception{});
+}
```
daniel-j-h: Can you clang-format with the .clang-format style file

libsweep/src/dummy.cc
```diff
@@ -34,6 +34,68 @@ void sweep_error_destruct(sweep_error_s error) {
   delete error;
 }
 
+// ------------------ Blocking Methods (blocking) ---------------- //
+
+void sweep_device_start_scanning(sweep_device_s device, sweep_error_s* error) {
+  SWEEP_ASSERT(device);
+  SWEEP_ASSERT(error);
+  SWEEP_ASSERT(!device->is_scanning);
+  (void)device;
```
daniel-j-h: You can remove this, we're using device below

libsweep/src/dummy.cc
```diff
@@ -34,6 +34,68 @@ void sweep_error_destruct(sweep_error_s error) {
...
+  if (device->is_scanning)
+    return;
+
+  device->is_scanning = true;
+}
+
+void sweep_device_stop_scanning(sweep_device_s device, sweep_error_s* error) {
+  SWEEP_ASSERT(device);
+  SWEEP_ASSERT(error);
+  (void)device;
```
daniel-j-h: remove

libsweep/src/dummy.cc
```diff
@@ -34,6 +34,68 @@ void sweep_error_destruct(sweep_error_s error) {
...
+  (void)error;
+
+  device->motor_speed = hz;
+}
+
+sweep_scan_s sweep_device_get_scan(sweep_device_s device, sweep_error_s* error) {
+  SWEEP_ASSERT(device);
+  SWEEP_ASSERT(error);
+  SWEEP_ASSERT(device->is_scanning);
+  (void)device;
```
daniel-j-h: remove

libsweep/src/dummy.cc
```diff
@@ -52,46 +114,72 @@ sweep_device_s sweep_device_construct(const char* port, int32_t bitrate, sweep_e
   return out;
 }
 
+void sweep_device_attempt_start_scanning(sweep_device_s device, sweep_error_s* error) {
+  SWEEP_ASSERT(device);
+  SWEEP_ASSERT(error);
+  SWEEP_ASSERT(!device->is_scanning);
+  (void)device;
```
daniel-j-h: remove

libsweep/src/protocol.cc
```diff
@@ -75,8 +78,10 @@ void write_command(serial::device_s serial, const uint8_t cmd[2], error_s* error
   packet.cmdByte2 = cmd[1];
   packet.cmdParamTerm = '\n';
 
-  serial::error_s serialerror = nullptr;
+  // pause for 2ms, so the device is never bombarded with back to back commands
+  std::this_thread::sleep_for(std::chrono::milliseconds(2));
```
daniel-j-h: Why is this a problem?

libsweep/src/sweep.cc
```diff
@@ -2,9 +2,61 @@
...
 #include <chrono>
+#include <condition_variable>
+#include <mutex>
+#include <queue>
 #include <thread>
 
+// A threadsafe-queue to store and retrieve scans
+class ScanQueue {
+public:
+  ScanQueue(int32_t max) : q(), m(), c() { max_size = max; }
```
daniel-j-h: What is q, m, c - can you be more explicit

libsweep/src/sweep.cc
```diff
@@ -2,9 +2,61 @@
...
+
+  ~ScanQueue() {}
```
daniel-j-h: Does nothing? remove

libsweep/src/sweep.cc
```diff
@@ -15,6 +67,8 @@ typedef struct sweep_error {
 typedef struct sweep_device {
   sweep::serial::device_s serial; // serial port communication
   bool is_scanning;
+  ScanQueue* scan_queue;
```
daniel-j-h: Why a raw pointer? Who's responsible for ownership here?

Either make ScanQueue default constructible or use a unique_ptr

libsweep/src/sweep.cc
```diff
@@ -100,33 +109,60 @@ void sweep_device_start_scanning(sweep_device_s device, sweep_error_s* error) {
...
+  int32_t speed = sweep_device_get_motor_speed(device, &speederror);
+  if (speederror) {
+    *error = sweep_error_construct("unable to start scanning. could not verify motor speed.");
+    sweep_error_destruct(speederror);
     return;
   }
+  // Check that the motor is not stationary
+  if (speed == 0 /*Hz*/) {
+    // If the motor is stationary, adjust it to default 5Hz
+    sweep_device_set_motor_speed(device, 5 /*Hz*/, &speederror);
```
daniel-j-h: Should we symbolize this default speed?

libsweep/src/sweep.cc
```diff
@@ -100,33 +109,60 @@ void sweep_device_start_scanning(sweep_device_s device, sweep_error_s* error) {
...
 
+  // Start SCAN WORKER
+  device->scan_queue->flush();
   device->is_scanning = true;
+  // START background worker thread
+  device->stop_thread = false;
+  // create a thread
+  std::thread th = std::thread(sweep_device_accumulate_scans, device);
+  // detach the thread so that it runs in the background and cleans itself up
+  th.detach();
```
daniel-j-h: Hmm, shouldn't we block (i.e. `.join()`) in the device's destructor? That is, set the stop scanning atomic flag, then wait until done.

libsweep/src/sweep.cc
```diff
@@ -169,7 +209,50 @@ void sweep_device_stop_scanning(sweep_device_s device, sweep_error_s* error) {
...
-sweep_scan_s sweep_device_get_scan(sweep_device_s device, sweep_error_s* error) {
+void sweep_device_wait_until_motor_ready(sweep_device_s device, sweep_error_s* error) {
+  SWEEP_ASSERT(device);
+  SWEEP_ASSERT(error);
+  SWEEP_ASSERT(!device->is_scanning);
+
+  sweep_error_s readyerror = nullptr;
+  int32_t motor_ready;
+  // Only check for 8 seconds (16 iterations with 500ms pause)
+  for (auto i = 0; i < 16; ++i) {
```
daniel-j-h: Why 16? Why 8 seconds?

libsweep/src/sweep.cc
```diff
@@ -169,7 +209,50 @@ void sweep_device_stop_scanning(sweep_device_s device, sweep_error_s* error) {
...
+    motor_ready = sweep_device_get_motor_ready(device, &readyerror);
+    if (readyerror) {
+      *error = readyerror;
+      return;
+    }
+    if (motor_ready == 0) {
+      return;
+    }
+    // pause for 500ms
+    std::this_thread::sleep_for(std::chrono::milliseconds(500));
```
daniel-j-h: Why?

sweeppy/sweeppy/__main__.py
```diff
@@ -16,9 +16,11 @@ def main():
         print('Motor Speed: {} Hz'.format(speed))
         print('Sample Rate: {} Hz'.format(rate))
 
+        # Starts scanning as soon as the motor is ready
         sweep.start_scanning()
 
-        # get_scans is coroutine-based generator lazily returning scans ad infinitum
+        # get_scans is coroutine-based generator lazily returning scans ad
+        # infinitum
```
daniel-j-h: Why break here?

libsweep/src/sweep.cc
```diff
@@ -399,6 +641,51 @@ void sweep_device_set_sample_rate(sweep_device_s device, int32_t hz, sweep_error
     sweep::protocol::error_destruct(protocolerror);
     return;
   }
+
+  // Check the status bytes do not indicate failure
+  const uint8_t status_bytes[2] = {response.cmdStatusByte1, response.cmdStatusByte2};
+  int32_t status_code = sweep::protocol::ascii_bytes_to_integral(status_bytes);
+  switch (status_code) {
+  case 11:
```
daniel-j-h: What is 11? Can we symbolize this

libsweep/README.md
dcyoung: The idea behind the non-blocking versions of certain methods, was to provide options for low-level development, where blocking or threading wasn't available.

libsweep/README.md
dcyoung: For non low-level development, the API should be unchanged. The "attempt_..." methods are provided as niche case alternatives... and not meant to be advertised. I went ahead and separated these methods out of the main README so that the distinction is clear. See if that makes more sense. 

If you have ideas for a better naming convention, please share! I'm not very content with the current "attempt_" and "direct_" pattern.

libsweep/include/protocol.h
dcyoung: That might make this PR look a lot more tidy.

libsweep/src/protocol.cc
dcyoung: The sweep device should be capable of processing a few commands back to back. But the response behavior can be unexpected in certain situations. I encountered occasional errors on devices when sending multiple commands rapidly back to back. We will continue to work on improving the device's capabilities, but for now, waiting a few milliseconds between successive commands results in a more robust SDK.

libsweep/src/sweep.cc
dcyoung: Yes. How would you like to do so? The API should concern itself with speeds and sample rates etc... but on the implementation side, I think it would be best if we operate in terms of protocol codes (speed setting code, sample rate setting code, error code, status byte code etc). 

I'm not sure what the best container for these will be. It might be nice in certain circumstances to be able to reference a "DEFAULT_MOTOR_SPEED_CODE" or something, but in other cases it might be useful to loop through a container to check for associated info related to specific code (ie: to get the sample rate in HZ from a sample rate code).

libsweep/src/sweep.cc
dcyoung: As I understand it, the detached background thread cleans itself up properly when it's method returns or terminates. Seeing how the behavior of the thread exists in the "background" with no need to synchronize, it seems like a much simpler solution to avoid explicitly managing the thread.

libsweep/src/sweep.cc
dcyoung: 11 is a failure status code that indicates an invalid parameter. These details are outlined in the protocol docs, but I think it might be best to include them in the protocol implementation. Perhaps making a container of possible status codes for each type of response?

libsweep/include/protocol.h
dcyoung: Correct, the SDK will no longer work with firmware v 1.0 which did not include the "MZ" command, among other things.

MikeGitb:
If this is to be released soon, may I suggest to not put the low level functions into the public API just yet?
As changing the API afterwards is usually undesirable I think it would better to have one or two iterations on it before settling for a design, but as there are probably only very few users of it it should imho not block the merge of this PR.

Talking about blocking the merge: For substantial changes like this I often take a triaging approach: Only hold of the merge, until the API is OK and the code appears to be bugfree (that is the tricky part) and then address implementation issues in separate (smaller) PRs.

dcyoung:
@MikeGitb I like the idea of waiting to add these "alternative" methods to the API until they are more fleshed out and the naming convention is settled. The rest of the API should remain unchanged. The implementation differs, and we've added new methods relevant to "motor ready", but the core start/stop methods etc are the same. I'll remove these from the API in the next few hours and update the PR.

dcyoung:
Reminder: Also need to update sweepjs bindings for the change in getMotorReady (now returns boolean, rather than a code)

libsweep/src/dummy.cc
```diff
@@ -34,6 +34,65 @@ void sweep_error_destruct(sweep_error_s error) {
   delete error;
 }
 
+// ------------------ Blocking Methods (blocking) ---------------- //
```
dcyoung: reorganize the methods to match sweep.h/sweep.cc and remove the outdated comments

sweepjs/README.md
```diff
@@ -42,10 +42,18 @@ sweep = new Sweep('/dev/ttyUSB0');
 sweep.startScanning();
 sweep.stopScanning();
 
+// waits until motor is ready
+sweep.waitUntilMotorReady();
+// code === 0 -> ready (motor ready), code !== 0 -> not ready
```
dcyoung: update sweepjs bindings reflect boolean return, instead of a motor ready code

sweeppy/README.md
```diff
@@ -62,10 +62,12 @@ class Sweep:
     def start_scanning(self) -> None
     def stop_scanning(self) -> None
 
-    def get_motor_speed(self) -> int
+    def wait_until_motor_ready(self) -> None
+    def get_motor_ready(self) -> int (0 indicates motor is ready)
```
dcyoung: indicate that a bool is returned

libsweep/README.md
dcyoung: I've gone ahead and removed the alternative methods from the API for now. We can merge this PR, and worry about the naming conventions etc later. The only changes to the public API will be the addition of two new methods (`get_motor_ready` and `wait_until_motor_ready`). As @MikeGitb  suggested, we can worked on the implementation details and naming conventions for the alternative bits AFTER this PR... and merge the updated API later when we're comfortable with it.

dcyoung:
Just pushed an update to the MAJOR version, and added a table for firmware compatibility in the README. If the appveyor + travis builds look good, I'm going to go ahead and merge this.

dcyoung:
Unfortunately, I realized after merging this that the "wait_until_motor_ready" method was not working properly when being called externally. It works fine internally. This might have something to do with the error handling. 

Luckily, this method is not actually necessary, given that "start_scanning" and "set_motor_speed" call it internally. Therefore, I stripped the method from the public API and quickly pushed the fix to master. I avoided re-basing master in case anyone already grabbed it. But my apologies if anyone has been working with it during the few hours before the fix.

Issue #69, opened 2017-04-13
lingering device with sweepjs bindings


dcyoung:
Something appears to be persisting after using the sweepjs bindings. Attempting to use an alternate method of interfacing with the device (ie: the sweep-visualizer) after running the sweepjs example requires that the device be reset. This is not the case with the sweeppy python bindings. My best guess is that the python bindings properly cleanup and and destroy the device on exit, while perhaps the sweepjs bindings do not?

daniel-j-h:
The Sweep device is always getting cleaned up by means of a ref counted ptr

- https://github.com/scanse/sweep-sdk/blob/8738b20e5d6752d3f339e52be6f637d137a84280/sweepjs/sweepjs.h#L45-L46
- https://github.com/scanse/sweep-sdk/blob/8738b20e5d6752d3f339e52be6f637d137a84280/sweepjs/sweepjs.cc#L41-L46

Ref counted because there might be a async. callback still in flight.
- https://github.com/scanse/sweep-sdk/blob/8738b20e5d6752d3f339e52be6f637d137a84280/sweepjs/sweepjs.cc#L188

dcyoung:
Hmmm.... Those links make sense to me, but #71 did not appear to fix the issue. 

Upon further inspection, it seems the example `index.js` was not terminating. Instead it was hung up somehow. This was preventing other applications from using the device. 

Firstly, I noticed that `index.js` does not explicitly call `stopScanning`. This shouldn't be an issue, as the destructor from libsweep should stop scanning, but it should be included for reference anyhow. 

I was seeing odd behavior when experimenting with `index.js`. If I placed a `stopScanning()` call after the `sweep.scan()`, it appeared to me that the callback passed in the `sweep.scan()` method was never being called, and the program hung indefinitely. With limited experimentation, I was able to get the example to print the samples, stopScanning, and terminate properly. https://github.com/scanse/sweep-sdk/commit/a3f955399db5b7fff7affd9c4867bb0272b64225

But I had to place the `stopScanning()` call inside the callback. See https://github.com/scanse/sweep-sdk/blob/a3f955399db5b7fff7affd9c4867bb0272b64225/sweepjs/index.js#L38

Obviously this isn't ideal. @daniel-j-h Do you have any ideas what might be causing this behavior.

daniel-j-h:
[...]

PR #71, opened 2017-04-16
Added close serial port to device_destruct in unix serial.cc

kent-williams:
Serial device was not being properly closed on unix systems. This is ok on windows.

### DIFF
--- a/libsweep/src/unix/serial.cc
+++ b/libsweep/src/unix/serial.cc
@@ -353,7 +353,11 @@ void device_destruct(device_s serial) {
 
   error_s ignore = nullptr;
   device_flush(serial, &ignore);
-  close(serial->fd);
+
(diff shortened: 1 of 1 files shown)


daniel-j-h:
Urgh. Good catch! Can you add an assertion like this one:

https://github.com/scanse/sweep-sdk/blob/6ed0e1d49b7fa8833f9e6951c500e34f9896eba0/libsweep/src/unix/serial.cc#L340-L342

as not being able to close the serial device will corrupt state - which will lead to the problem you saw here. Then this is ready to go in. Thank you Kent.

daniel-j-h:
Also this is related to https://github.com/scanse/sweep-sdk/issues/69, no? If so feel free to merge and close the issue.

Issue #73, opened 2017-04-27
Delay in serial buffering of DX response on RaspberryPi, corrupts subsequent responses


dcyoung:
I've discovered an issue but not entirely sure what is causing it.

- OS: linux
- Platform: so far, I can only reproduce this behavior on the raspberryPi
- Library: libsweep (and by extension sweepy + sweepjs as well)
- Repeatable: around 25% of the time

Code to reproduce:
```c++
sweep::sweep device{argv[1]};
device.set_sample_rate(500); // errors here, description below
```

The expected behavior should be as follows:

- During device creation, `sweep_device_stop_scanning()` is called internally... which performs the following:
  - sends `DX\n` command to stop any potential active data stream
  - waits 35ms to allow any incoming data blocks, and the DX response (`DX00P\n`) to come in
  - flushes buffer (including the first DX response)
  - sends another `DX\n` command
  - reads the DX response (`DX00P\n`) that should be responding to the 2nd command
- At this point the serial buffer should be empty!
- Set sample rate
  - `sweep_device_set_sample_rate()` sends `LR01\n` command to set the sample rate and then reads LR response (`LR01\n00P\n`)

The issue is a checksum error when attempting to read the LR response. Instead of `LR01\n00P\n`, the response that is read from the serial buffer is `DX00P\nLR0`. It appears that a `DX - Stop Scanning` response (`DX00P\n`)  is being left in the serial buffer, and is therefore being interpreted as the beginning of the LR response. 

Using a protocol analyzer I have confirmed that the timing of the commands and responses, as well as their byte content, is as expected:
Here is a zoomed out timeline:
- device is constructed (sends 2 DX commands 35 ms apart, both receive valid responses)
- Set sample rate (with valid response)
- program errors and the device is destructed (sends 2 DX commands 35 ms apart, both receive valid responses)

![sniffer](https://cloud.githubusercontent.com/assets/7978433/25506153/1b97b722-2b5a-11e7-8e72-c7f459442e10.png)

From this I can tell that the communication over the wire is fine. No extra "DX" response (`DX00P\n`) is coming in, and no response is delayed. The bytes from the "LR" response (`LR01\n00P\n`) are valid as well. 

It seems then that somehow the buffer isn't being flushed properly. @daniel-j-h Any idea what could be causing this?

Edit: it appears the flush is NOT the issue. See below.

dcyoung:
[...]

PR #74, opened 2017-04-28
Temp Fix for Delayed Buffering of Stop Response during Device Construction on Raspberry Pi

dcyoung:
This a temporary solution to the problems described in #73. See the issue for details.

It would be nice to know the root cause of the bug, but this temporary fix will allow development of the 3d-scanner to commence.

### DIFF
--- a/libsweep/src/sweep.cc
+++ b/libsweep/src/sweep.cc
@@ -294,6 +294,7 @@ void sweep_device_stop_scanning(sweep_device_s device, sweep_error_s* error) try
   } catch (const std::exception& ignore) {
     // Catch and ignore the error in the case of the stop response containing garbage bytes
     // Occurs if the device was actively streaming data before the stop cmd
+    (void)ignore;
   }
 
   // Flush any bytes left over, in case device was actively streaming data before the stop cmd
--- a/libsweep/src/sweep.cc
+++ b/libsweep/src/sweep.cc
@@ -279,14 +279,27 @@ void sweep_device_stop_scanning(sweep_device_s device, sweep_error_s* error) try
 
   sweep::protocol::write_command(device->serial, sweep::protocol::DATA_ACQUISITION_STOP);
 
-  // Wait some time, for the device to register the stop cmd and stop sending data blocks
+  // Wait some time for a few reasons:
+  // 1. allow time for the device to register the stop cmd and stop sending data blocks
+  // 2. allow time for slower performing devices (RaspberryPi) to buffer the stop response
+  // 3. allow a grace period for the device to perform its logging routine before sending a second stop command
   std::this_thread::sleep_for(std::chrono::milliseconds(35));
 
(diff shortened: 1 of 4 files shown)


libsweep/src/sweep.cc
```diff
@@ -218,11 +218,25 @@ void sweep_device_stop_scanning(sweep_device_s device, sweep_error_s* error) {
     return;
   }
 
-  // Wait some time, for the device to register the stop cmd and stop sending data blocks
+  // Wait some time for a few reasons:
+  // 1. allow time for the device to register the stop cmd and stop sending data blocks
+  // 2. allow time for slower performing devices (RaspberryPi) to buffer the stop response
+  // 3. allow a grace period for the device to perform its logging routine before sending a second stop command
```
daniel-j-h: As said in the ticket, if we can just make the timeout below a bit longer I would go for that. Even hundreds of milliseconds are fine for construction if it makes things easier for us.

dcyoung: Check [ticket](https://github.com/scanse/sweep-sdk/issues/73) for response.

dcyoung:
Seems that this is definitely an issue other Raspberry Pi's as well, and that this fix is resolving the issue. See #88 

@daniel-j-h Perhaps there is a better fix applicable to `unix/serial.cc`. I didn't write that code, but if you have any ideas for a possible source or fix let me know and I'd be happy to try and debug the `unix/serial.cc`.

In the meantime though, I'd like to get this merged as I think the Pi might be a common use case.  Is there any reason we shouldn't merge this now, and bring a better fix in a later PR?

daniel-j-h:
Urgh - great that it's fixed upstream. Feel free to merge if it's needed until Raspbian picks the fix up.

But please ticket this for later removal and maybe link it even from the source code. I don't want to add too many hacks for platform quirks.

There _never_ should be a downside of doing this, right? Otherwise we need to detect the Raspberry env. with CMake and only conditionally enable the workaround.

daniel-j-h:
@dcyoung still want to bring this in? I just added some release docs and a changelog [here](https://github.com/scanse/sweep-sdk/pull/97). Feel free to merge the release docs pull request first and then make a release with these Raspberry specific changes.

dcyoung:
@daniel-j-h  sounds good. I was dragging my feet a little, waiting to get a better estimate on the release timeframe for the Raspbian build.... but I think its silly to wait on it. The rPi will be a relatively common use case and I'd prefer to avoid confusion for those who aren't aware of this branch. 

As for the release docs, I made some comments there. Let me know what you think, and if you want to keep things simple I'll merge them in that order. Additionally, should we merge #98 (API cleanup PR) later, or before all these others? I'm wondering if the code shuffling will make rebasing all the other PRs more cumbersome.

daniel-j-h:
I think order does not really matter as there probably will be no conflicts.

PR #75, opened 2017-04-28
Stop scanning after retrieving scan in SweepJS example

dcyoung:
This is a temporary fix to the problems described in issue #69. See issue for details.

### DIFF
--- a/sweepjs/index.js
+++ b/sweepjs/index.js
@@ -19,17 +19,23 @@ if (require.main === module) {
 
   console.log('Starting data acquisition as soon as motor is ready...');
   sweep.startScanning();
-
+  
+  console.log('Retrieving 1 scan...');
   sweep.scan(function (err, samples) {
     if (err) {
-      return console.log(err);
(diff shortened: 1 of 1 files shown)


daniel-j-h:
I don't think the stop scanning call here is correct - you're placing it inside the scan callback, which means after the first full scan we shut down. I really don't want to advertise the stop scanning function in the Js examples as it's super easy to get this wrong wrt. async. functions: if you place the stop scanning call at the end, it will be called immediately because the scan function is async.

dcyoung:
Hmmm. This was only meant to be a temporary fix, as the sweepjs example is currently broken for me. But you are right that including the stop_scanning in the callback is rather poor practice, and could lead to issues in certain setups. I'm not sure why the application was hanging in its absence or how to properly remedy the behavior. Any ideas?

dcyoung:
Closing this for now, as it is not the appropriate fix. Will try to test the issue again next week and see if I can't find a better solution.

PR #78, opened 2017-04-30
Refactors Queue

daniel-j-h:
- puts queue impl. into `queue.h`
- makes it generic on its value type
- marks thread safe members `mutable`
- no need for a `unique_ptr<queue>`, just store by value

### DIFF
--- a/libsweep/include/queue.h
+++ b/libsweep/include/queue.h
@@ -1,0 +1,68 @@
+#ifndef SWEEP_QUEUE_62C8F42E8DD5_H
+#define SWEEP_QUEUE_62C8F42E8DD5_H
+
+/*
+ * Thread-safe queue.
+ * Implementation detail; not exported.
+ */
+
+#include <stdint.h>
+
+#include <atomic>
+#include <condition_variable>
+#include <mutex>
+#include <queue>
+#include <utility>
+
+namespace sweep {
+namespace queue {
+
+template <typename T> class queue {
+public:
+  queue(int32_t max) : max_size(max) {}
+
+  // Empty the queue
+  void clear() {
+    std::unique_lock<std::mutex> lock(the_mutex);
+    while (!the_queue.empty()) {
+      the_queue.pop();
+    }
+  }
+
+  // Add an element to the queue.
+  void enqueue(T v) {
+    std::lock_guard<std::mutex> lock(the_mutex);
+
+    // if necessary, remove the oldest scan to make room for new
+    if (static_cast<int32_t>(the_queue.size()) >= max_size)
+      the_queue.pop();
+
+    the_queue.push(v);
+    the_cond_var.notify_one();
+  }
+
+  // If the queue is empty, wait till an element is avaiable.
+  T dequeue() {
+    std::unique_lock<std::mutex> lock(the_mutex);
+    // wait until queue is not empty
+    while (the_queue.empty()) {
+      // the_cond_var could wake up the thread spontaneously, even if the queue is still empty...
+      // so put this wakeup inside a while loop, such that the empty check is performed whenever it wakes up
+      the_cond_var.wait(lock); // release lock as long as the wait and reaquire it afterwards.
+    }
+    auto v = the_queue.front();
+    the_queue.pop();
+    return v;
+  }
+
+private:
+  int32_t max_size;
+  std::queue<T> the_queue;
+  mutable std::mutex the_mutex;
+  mutable std::condition_variable the_cond_var;
+};
+
+} // ns queue
+} // ns sweep
+
+#endif
--- a/libsweep/src/sweep.cc
+++ b/libsweep/src/sweep.cc
@@ -1,59 +1,10 @@
 #include "sweep.h"
 #include "protocol.h"
+#include "queue.h"
 #include "serial.h"
 
-#include <atomic>
 #include <chrono>
-#include <condition_variable>
-#include <mutex>
-#include <queue>
 #include <thread>
-
-// A threadsafe-queue to store and retrieve scans
-class ScanQueue {
-public:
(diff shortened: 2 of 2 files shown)


libsweep/include/queue.h
```diff
@@ -0,0 +1,68 @@
...
+template <typename T> class queue {
+public:
+  queue(int32_t max) : max_size(max), the_queue(), the_mutex(), the_cond_var() {}
+
+  // Empty the queue
+  void flush() {
+    std::unique_lock<std::mutex> lock(the_mutex);
+    while (!the_queue.empty()) {
+      the_queue.pop();
+    }
```
MikeGitb: Why not use `clear()`

libsweep/include/queue.h
```diff
@@ -0,0 +1,68 @@
...
+
+  // If the queue is empty, wait till an element is avaiable.
+  T dequeue() {
+    std::unique_lock<std::mutex> lock(the_mutex);
+    // wait until queue is not empty
+    while (the_queue.empty()) {
+      // the_cond_var could wake up the thread spontaneously, even if the queue is still empty...
+      // so put this wakeup inside a while loop, such that the empty check is performed whenever it wakes up
+      the_cond_var.wait(lock); // release lock as long as the wait and reaquire it afterwards.
+    }
```
MikeGitb: There is an overload for `.wait` that takes a predicate which could shorten this a bit:
http://en.cppreference.com/w/cpp/thread/condition_variable/wait

    the_cond_var.wait(lock, [this]{ return !this->the_queue.empty(); } );

Whether this would make the code actually more readable is up for debate

libsweep/include/queue.h
```diff
@@ -0,0 +1,68 @@
...
+#include <mutex>
+#include <queue>
+#include <utility>
+
+namespace sweep {
+namespace queue {
+
+template <typename T> class queue {
+public:
+  queue(int32_t max) : max_size(max), the_queue(), the_mutex(), the_cond_var() {}
```
MikeGitb: Explicitly calling the default constructors of `the_queue(), the_mutex(), the_cond_var()` is not necessary.

libsweep/include/queue.h
```diff
@@ -0,0 +1,68 @@
...
+  // If the queue is empty, wait till an element is avaiable.
+  T dequeue() {
+    std::unique_lock<std::mutex> lock(the_mutex);
+    // wait until queue is not empty
+    while (the_queue.empty()) {
+      // the_cond_var could wake up the thread spontaneously, even if the queue is still empty...
+      // so put this wakeup inside a while loop, such that the empty check is performed whenever it wakes up
+      the_cond_var.wait(lock); // release lock as long as the wait and reaquire it afterwards.
+    }
+    auto v = the_queue.front();
```
MikeGitb: probably would use `std::move` here

libsweep/src/sweep.cc
```diff
@@ -65,8 +16,8 @@ typedef struct sweep_error {
 typedef struct sweep_device {
   sweep::serial::device_s serial; // serial port communication
   bool is_scanning;
-  std::unique_ptr<ScanQueue> scan_queue;
-  std::atomic<bool> stop_thread;
+  mutable sweep::queue::queue<sweep_scan_s> scan_queue;
+  mutable std::atomic<bool> stop_thread;
```
MikeGitb: Is there a reason, why those should be mutable?

daniel-j-h:
@MikeGitb the code is more or less moved from the `sweep.cc` file as-is. Not rvalue friendly in both the enqueue as well as the dequeue member functions, yep.

mutable for signalling already-thread-safe:
https://channel9.msdn.com/posts/C-and-Beyond-2012-Herb-Sutter-You-dont-know-blank-and-blank

MikeGitb:
@daniel-j-h: I've seen the presentation and with all due respect to Herb Sutter: He went a little overboard with one or two examples there. 
He is certainly right that calling const member functions should never introduce data races.
However, const-correctness is still about more than just thread-safety (otherwise you wouldn't need it in singlethreaded programs at all) and the old guideline that calling a const memberfunction should not change the observable state of the object is still important.

The reason why things like `std::mutex`  and other synchronization primitives are often declared mutable is because not because they CAN me modified in a const memberfunction without violating THE NEW rule, but because

1. They often NEED to be modified in a const member function to ensure the aforementioned thread safety. The only alternative would be to declare a member function that is logically const (like getting the size of a container) non-const.
2. Their state is not visible externally
3. Even if it is visilbe, their state is often the same at the beginning and end of a function

On the other hand, there are examples, where you have to declare members mutable that aren't even threadsafe in the frist place (e.g. a cache) and you are ensuring threadsafety via an extra synchronization primitive.

So imho if one wants to document that a object is thread safe, this should either happen by naming the class accordingly or via comments -- not via a language construct with a related, but different purpose.

dcyoung:
I'll build this now and give it a proper review. Should we aim to change how we specify queue size as well, perhaps another version of `start_scanning(int max_queue_size)` that allows users to specify the max queue size? Or should we leave that for a later PR?

libsweep/include/queue.h
dcyoung: I was under the impression clear does not work with std::queue... or did you mean to use the method name `clear` as opposed to `flush`. I agree that is probably a better naming choice.

dcyoung:
Looks better than what we had previously (crammed into sweep.cc)

libsweep/include/queue.h
MikeGitb: Sorry, my mistake - Apparently I mixed up `std::queue` with `std::deque`.

Yes, I'd try to stick with the stl naming convention in general (also for the other function names), but as this is only an internal datastructure anyway there is no rush.

daniel-j-h:
Adapted to suggestions, thanks!

PR #82, opened 2017-05-04
Cpp-fy libsweep/protocol

MikeGitb:
A couple of small refactorings that make protocol.h and protocol.cc a bit more "cpp-ish"

### DIFF
--- a/libsweep/include/protocol.h
+++ b/libsweep/include/protocol.h
@@ -194,7 +194,7 @@ inline void integral_to_ascii_bytes(const int32_t integral, uint8_t bytes[2]) {
   SWEEP_ASSERT(bytes);
 
   // Numbers begin at ASCII code point 48
-  const uint8_t ASCIINumberBlockOffset = 48;
+  const uint8_t ASCIINumberBlockOffset = '0';
 
   const uint8_t num1 = (integral / 10) + ASCIINumberBlockOffset;
   const uint8_t num2 = (integral % 10) + ASCIINumberBlockOffset;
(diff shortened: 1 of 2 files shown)


MikeGitb:
Case of bad timing, as it probably conflicts with #80.
If merging turns out to be non-trivial, I'd be happy to rebase onto the result of #80

libsweep/include/protocol.h
```diff
@@ -6,7 +6,7 @@
  * Implementation detail; not exported.
  */
 
-#include <stdint.h>
+#include <cstdint>
```
daniel-j-h: I never fully understood the difference here - I thought if we want to use `int32_t` (unprefixed) we have to include `stdint.h`, and if we include `cstdint` we _have to_ use `std::int32_t` to be on the safe side?

libsweep/include/protocol.h
```diff
@@ -206,8 +206,7 @@ inline void integral_to_ascii_bytes(const int32_t integral, uint8_t bytes[2]) {
 inline int32_t ascii_bytes_to_integral(const uint8_t bytes[2]) {
   SWEEP_ASSERT(bytes);
 
-  // Numbers begin at ASCII code point 48
-  const uint8_t ASCIINumberBlockOffset = 48;
+  const uint8_t ASCIINumberBlockOffset = '0';
```
daniel-j-h: Line 197 does the same and should probably adapted, too.

libsweep/include/protocol.h
MikeGitb: @daniel-j-h: I think you are right <cstdint> only guarantees the availability of `std::int32_t` not the un-prefixed version.

libsweep/include/protocol.h
MikeGitb: Right, overlooked that.

MikeGitb:
Ok, I think I fixed your comments: I left the stdint.h header in place and removed  the two other ones that appered to be no longer in use anyway.
 
Regarding the merge order, do whatever you think is easiest. when in doubt I'd give priority to #80 as this one here makes mostly cosmetic changes. It was actually meant to be the first step of multiple PRs (ironically also introducing exceptions).

PR #84, opened 2017-05-05
Added templates for contribution, issues and pull requests

dcyoung:
This PR adds templates for new issues and pull requests. It places templates in a `.github/` directory at the root of repo to avoid clutter. The templates provide general guidelines for contribution. The existing documentation remains untouched.

This was really just a start. We will likely see a rapid pickup in community contribution as backers begin to receive devices. I am open to completely reformatting these templates to provide contributors with necessary information, or to better clarify our policies.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,0 +1,29 @@
+# Help 
+The [issues](https://github.com/scanse/sweep-sdk/issues) section is for bug reports and feature requests. If you need help with a project, please use the [community](http://community.scanse.io/) forum.
+
+---
+# Bugs
+#### Before reporting a bug
+
+1. Search [issue tracker](https://github.com/scanse/sweep-sdk/issues) for similar issues.
+2. Make sure you are using the latest version of the library and that you have thoroughly reviewed all provided documentation.
+
+#### How to report a bug
+
+1. Specify the version of the library in use when the bug occurred.
+2. Specify your operating system, platform and any other details you think might be relevant.
+3. Describe the problem in detail. Explain what happened, and what you expected would happen.
+4. If possible, provide a small code-snippet or test-case that reproduces the issue.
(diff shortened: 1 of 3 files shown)


daniel-j-h:
I like the contributing guidelines! 👍

Not sure about the templates for issues and pull request though - it feels rather mechanical. What about minimizing the content for the issue and pull request template and putting hints for the user into comments? For example

```
<!--
Please specify the sweep-sdk version, your device's firmware, operating system, and be as specific as possible.
-->
```

The user will then see it but we don't clutter every issue and pull request with a form-like header.

What do you think?

dcyoung:
Hmmm, although it is definitely "mechanical", the templates standardize the bug reports. I think this will help in the long run. Posters will inevitably forget to include this kind of information, and we'll have to ask for it in the first reply. The rest could be comments/suggestions.  Not sure what we lose by presenting a methodical template. 

Having said that, I agree that the issues template is a bit verbose. Perhaps we could do the following: 
Make the guidelines for each section a simple comment, but leave the title as rendered markdown. Instead of including the checklist in the comments, we'd probably replace the checklists with simple descriptions.  So a section of the template might look like:

```
#### Description:
<!--
 Describe the bug or feature request in detail. 
 A code snippet, screenshot, and small-test help us understand.
 Explain what happened and what you expected to happen.
-->
```

As for the PR template, those are really just guidelines... so we can put those in a comment so it doesn't render. Although it does already include a simple note: `Once you have reviewed the provided guidelines, you can delete this text so it does not clutter the pull request.`

dcyoung:
I went ahead and updated the issue template so that it provides a strong template, but does not clutter the issue with extra markdown.

daniel-j-h:
Perfect thank you, this is great! Should we do the same for the pull request template?

Feel free to merge if you think it's ready!

dcyoung:
Terrific! I'll clean up the PR template a bit and merge today.

PR #87, opened 2017-05-08
[libsweep] Forward errors from accumulate_scans to get_scan

MikeGitb:
#### Scope of changes
Forward errors from accumulate_scans to get_scan.

Also adds an additional function that allows to cancel a read from a different thread, but doesn't expose this in the API

#### Known Limitations
Doesn't address the more fundamental question, if an error inside  accumulate_scans should actually stop the worker thread and how the user is supposed to restart it.

### DIFF
--- a/libsweep/include/queue.hpp
+++ b/libsweep/include/queue.hpp
@@ -10,12 +10,17 @@
 
 #include <atomic>
 #include <condition_variable>
+#include <exception>
 #include <mutex>
 #include <queue>
 #include <utility>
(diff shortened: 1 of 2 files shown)


MikeGitb:
Closing this for now - I think the worker thread should handle errors gracefully and not just die on the user

dcyoung:
@MikeGitb I can't seem to find your contact info through the git profile. Could you shoot us an email at team@scanse.io

Issue #89, opened 2017-05-11
Sensor protocol description nitpicking


MikeGitb:
The [protocol description file](https://github.com/scanse/sweep-sdk/blob/master/libsweep/doc/serial_protocol_spec.md) e.g. says

> Azimuth: Angle that ranging was recorded at (in degrees). Azimuth is a float value, transmitted as a 16 bit int. This needs to be converted from 16bit int to float. Use instructions in the Appendix. Note: the lower order byte is received first, higher order byte is received second.

I'm not sure if this is just confusing or even wrong, but in any case wouldn't it be simpler to say something like:

> Azimuth: Angle that ranging was recorded at (in degrees). Azimuth is transmitted as a 16 bit **fixed point** value with a scaling factor of 16 (4bit after radix) in little endian format (first byte is low order).

Also the conversion from wire format to storage format seems to be unnecessarily complicated. Wouldn't a simple

    static_cast<int>(angle/16.0 * 1000.0) 

do the trick?
Or have I overlooked something?

dcyoung:
I like the simplification a lot! That could help clear up a fair amount of confusion. I'll check with @kent-williams  to see if there is any edge case or reason why we're using the more complicated formula.  Otherwise I'll whip this up into a PR now.

PS: @MikeGitb we've been trying to get a hold of you directly. Please email `team@scanse.io` when you get a chance.

PR #90, opened 2017-05-11
Simplify azimuth conversion

dcyoung:
#### Scope of changes
Simplifies the conversion of the azimuth field. See #89 
Removes the overly complicated conversion method (`u16_to_f32`) from the protocol header, and updates the protocol docs.

#### Known Limitations 
Perhaps the docs should also include example conversions in languages other than js... as that really isn't the most likely use case. It was chosen as the example language more for simplicity and readability (almost like pseudo code).  But most users will likely be using c++ or something similar.


MikeGitb:
I think `u16_to_f32` is a misnomer  (and actually was one of the reasons I got confused initially). As far as I can tell that function is specific to the conversion of the angle from wire representation to float, so why not call it accordingly, like e.g. either 
- `int angle_raw_to_millideg(uint16_t raw)`
- `float angle_raw_to_deg(uint16_traw)`

The first one has the advantage, that it already produces the result in the right format, so there isn't a double conversion, the latter is the more generic

MikeGitb:
OK, never mind, I can make that PR myself

dcyoung:
@MikeGitb  Completely agree!

Issue #91, opened 2017-05-14
Error handling in worker thread


MikeGitb:
I finally have some time and thought about tackeling the question of what should happen when an error occures in the worker thread (currently the thread will simply stop). But before I create a PR I wanted to discuss my ideas first:

- First of all I think we need to distinguish between faults from which the thread should try to recover (e.g. crc errors) and error from which recovery is infeasible (device got disconnected).

- In case of recoverable errors I'd just try to find the next valid packet and give up after a few iterations.

- In case of non-recoverable errors, the loop would exit and the error would be propagated through the queue to `sweep_device_get_scan`.  

The question is what the user or the library is supposed to to to bring the system back into a defined state (we could for example automatically call `sweep_device_stop_scanning` or require this from the user)

dcyoung:
> In case of recoverable errors I'd just try to find the next valid packet and give up after a few iteration

Currently `libsweep` does not perform any kind of intelligent parsing. Instead it relies on the constant assumption that the data-block bytes arrive in order and without interruption until the stream is stopped. Technically, because each byte can take any form, it isn't possible to guarantee that you can find the next valid packet after an error. Consider the case of a missing byte, where the first byte of the proceeding data block is incorrectly included in the interpretation of the last byte of the current data block. When trying to find the next valid data block, you could slide a window (7 bytes long) across the incoming stream, checking if the values make sense.... but it wouldn't be a guarantee. Granted if the following check out, then it is very unlikely due to random chance:
1. validating the checksum
2. validating angle falls between [0:360] 

I'm thinking of all the possible non-recoverable errors that would halt the process. In every case it seems necessary to `stop_scanning` and potentially even `reset`. I don't think we should require that the user handle stopping the scanning process. 

I'm also wondering if it would be worth it to try and reset/restart the scanning automatically such that the user just sees a large gap in data while the sensor resets and starts scanning again. But this might require we keep track of state info (like sample rate) which the device forgets across power cycles. For now it seems better to just shutdown and report gracefully. Forcing graceful shutdown should also help us track down the root of the errors that occur as users report them.

MikeGitb:
If I understand correctly, this project is being shut down so I'm closing this issue.

PR #93, opened 2017-05-16
Error propagation

MikeGitb:
### Scope of changes

Propagates exceptions occuring in the worker thread to `sweep_device_get_scan`.

- Modifies queue code to play nice with move-only types
- Manages dynamically allocated memory via unqiue_ptr as long as possible to prevent any memory leaks
- Combines pointer to scan and exception pointer (one of them being nullptr) into one struct and passes them through the queue
- Removes some redundant type defs accross the code base for consistency reasons (ccan drop that commit from the PR if prefered)

### Known Limitations

- Not yet tested (seeking early feedback)
- Actual error handling (instead of just reporting) might be desired (see https://github.com/scanse/sweep-sdk/issues/91)

### DIFF
--- a/libsweep/src/sweep.cc
+++ b/libsweep/src/sweep.cc
@@ -6,20 +6,16 @@
 #include "sweep.h"
 
 #include <chrono>
+#include <exception>
 #include <string>
 #include <thread>
 
 int32_t sweep_get_version(void) { return SWEEP_VERSION; }
 bool sweep_is_abi_compatible(void) { return sweep_get_version() >> 16u == SWEEP_VERSION_MAJOR; }
 
-struct sweep_error { std::string what; };
-
-struct sweep_device {
-  sweep::serial::device_s serial; // serial port communication
-  bool is_scanning;
-  sweep::queue::queue<sweep_scan_s> scan_queue;
-  std::atomic<bool> stop_thread;
-} ;
+struct sweep_error {
+  std::string what;
+};
 
 #define SWEEP_MAX_SAMPLES 4096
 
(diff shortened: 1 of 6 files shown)


dcyoung:
builds and runs just fine for me under normal conditions, but I haven't had a chance to test the actual error propagation yet. I'll try to induce  or simulate some errors tomorrow and check that everything works.

dcyoung:
Sorry for the delay. I wasn't able to get a device to error (that's a good thing in most cases haha ), but manually throwing an exception in accumulate scans definitely propagates through. So its looking good to me. @daniel-j-h any comments? if not I'll merge this in.

libsweep/src/sweep.cc
```diff
@@ -6,29 +6,38 @@
...
+
+struct sweep_device {
+  sweep::serial::device_s serial; // serial port communication
+  bool is_scanning;
+  std::atomic<bool> stop_thread;
+
+  struct Element {
+    std::unique_ptr<sweep_scan> scan;
+    std::exception_ptr error;
+  };
```
daniel-j-h: 👍 great idea - do we need the `unique_ptr` here or can we just have a default constructed `sweep_scan` object? Or more concretely a scan object with `count=0`.

MikeGitb: currentl, a `sweep_scan` object is a struct of size ~12KB irrespective of how many actual scans it contains, so you certainly don't want it be copied around more than absolutely necessary. I actually meant to replace the current implementation of `sweep_scan` by a vector (essentaillly the same as the representation in the public c+ api), but I didn't want to mix up those two changes in a single PR.

daniel-j-h: Oh I see we still store fixed size arrays in the scan

dcyoung: shall we go ahead and merge this then? @MikeGitb I like the sound of that future PR

daniel-j-h: 👍

dcyoung:
@MikeGitb can you rebase/fix merge conflicts on master and I'll merge this.

MikeGitb:
done

dcyoung:
Merged. Thanks!

PR #97, opened 2017-05-24
Adds release documentation

daniel-j-h:
Documenting the release process.

I also like to keep a copy of the changelog in the repo. and not only [on Github](https://github.com/scanse/sweep-sdk/releases).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,33 @@
+# Changelog
+
+
+## v1.1.1
+
+This release is compatible with device firmware v1.2.
+
+Changes:
+- libsweep: Adapts motor calibration timeout to firmware changes
+
+
+## v1.1.0
+
+This release is compatible with device firmware v1.1.
+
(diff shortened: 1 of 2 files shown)


dcyoung:
I'd definitely like for us to have a changelog in the repo as well. Any interest in using an automated process to generate the log? Something like [https://github.com/skywinder/github-changelog-generator](https://github.com/skywinder/github-changelog-generator). For best results we'd likely have to be more diligent about how we title and close issues/PRs.

If not, I'm fine to do the log by hand. But should we adopt a consistent format for each release (eg: `Features`, `Improvements`, `Bugfixes`, `Known Issues` etc), or leave it unstructured? If you'd prefer to leave it as is, that's fine too.  Just throwing some ideas out there.

Seems travis is having some issues (but I think its unrelated to the change).

daniel-j-h:
I don't think we will have releases with more than a few lines of changelog notes :)

PR #98, opened 2017-05-24
Hides impl. details from API surface

daniel-j-h:
Features not exposed via `SWEEP_API` must not be exposed in the interface header.
This changeset moves some code around to cleanup the API surface.


dcyoung:
Thanks! I'd been meaning to do this as well. I'll give it a once over today to make sure nothing got dropped in the shuffling. Any idea why the Travis build is failing? At a glance, the output didn't seem to point anywhere in the code.

dcyoung:
I went ahead removed the `sweep_device_get_scan_direct()` as it is now hidden, and unused. I also cleaned up some comments etc. and edited `dummy.cc` to match `sweep.cc`. Otherwise looks good. Go ahead and merge.

Issue #102, opened 2017-06-14
libsweep does not handle unexpected unplug gracefully


sevenbitbyte:
Root cause of: https://github.com/scanse/sweep-ros/issues/11

#### libsweep version + affected bindings
This is an issue in the libsweep software package

#### operating system
I've only tested on Linux, current Windows behavior in the same situation is unknown

#### Description:

I have determined the root cause of the issue to be here: https://github.com/scanse/sweep-sdk/blob/master/libsweep/src/unix/serial.cc#L332

When the device is unplugged this results in an EOF event where all calls to `wait_readable` will return true but calls to `read` returns zero bytes. This sends us into an infinite loop as this specific failure mode is not addressed at all in the loop so we're stuck here forever.

#### Repro steps

- Plugin sensor
- Start a system monitor: `htop`
- Start sweep_node `roslaunch sweep_ros sweep.launch`
- Wait until scans are collected(1-5s): `rostopic hz /pc2`
- Unplug sensor

Observed result
- CPU usage goes from ~0.1% to over 90%
- No error messages in /rosout or log files
- Node stops publishing to `/pc2` topic

Expected result
- Error message indicating some form of sensor failure/read timeout
- CPU usage unchanged

sevenbitbyte:
Examining other failure modes in libsweep it appears the right thing here is to throw an exception(possibly causing the calling process to terminate abruptly). I will produce a patched version to bring this failure mode inline with the behavior I see employed elsewhere.

PR #103, opened 2017-06-14
Unplug exception

sevenbitbyte:
#### Issues

When the cable becomes unplugged the device_read function goes into an infinite loop causing the caller to hang and the process to consume 100% CPU cycles. This failure mode can be more gracefully handled by detecting EOF.

* https://github.com/scanse/sweep-ros/issues/11
* https://github.com/scanse/sweep-sdk/issues/102

#### Scope of changes

Modified Linux serial support to throw an exception if an EOF condition is detected.

#### Known Limitations

The error handling facilities are not very graceful and if the caller does not handle the exception, it result in a crash in upstream programs which may not have previously been expected(but neither was a hang). I have not yet been able to fully recover the connection after handling the exception but now you atleast get a clear error message.

### DIFF
--- a/libsweep/src/unix/serial.cc
+++ b/libsweep/src/unix/serial.cc
@@ -254,17 +254,21 @@ device_s device_construct(const char* port, int32_t bitrate) {
 
   int32_t fd = open(port, O_RDWR | O_NOCTTY | O_NONBLOCK);
 
+  printf("device_construct - 1\n");
   if (fd == -1)
     throw error{"opening serial port failed"};
 
+  printf("device_construct - 2\n");
   if (!isatty(fd))
     throw error{"serial port is not a TTY"};
 
   struct termios options;
 
+  printf("device_construct - 3\n");
   if (tcgetattr(fd, &options) == -1)
     throw error{"querying terminal options failed"};
 
+  printf("device_construct - 4\n");
   // Input Flags
   options.c_iflag &= ~(INLCR | IGNCR | ICRNL | IGNBRK);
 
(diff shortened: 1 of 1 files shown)


sevenbitbyte:
I've code reviewed the windows implementation and believe that this failure mode will result in an exception on windows(nearly same as how my patch works) so no windows patch should be needed. But if someone can verify that would be great.

dcyoung:
Thanks for the catch! I'm not sure if the perror() is necessary, but otherwise this looks good to me. @daniel-j-h any reason not to merge this ASAP?

I've tested on windows, and indeed an exception is thrown...  `Error: reading from serial device failed`.

sevenbitbyte:
I agree the perror _should_ not be required but from the ROS support layer I was not getting any error messages without doing it and I didn't fully understand your error bubbling approach

sevenbitbyte:
Is there a way we can get these exception messages to get printed to standard error when used under ROS?

MikeGitb:
@sevenbitbyte: My mistake. I forgot to upstream a fix for this. 

https://github.com/scanse/sweep-sdk/pull/105 should fix this

sevenbitbyte:
Ah makes sense.

dcyoung:
Merged! Thanks @sevenbitbyte

libsweep/src/unix/serial.cc
```diff
@@ -338,6 +338,8 @@ void device_read(device_s serial, void* to, int32_t len) {
         } else {
           throw error{"reading from serial device failed"};
         }
+      } else if(ret == 0){
+        throw error{"encountered EOF on serial device"};
```
daniel-j-h: Sorry I'm a bit late to this, but can there be situations where we don't encounter EOF and a intermediate read returns 0? Or to be more precise: should we check `feof(serial->fd)` instead?

PR #105, opened 2017-06-15
[libsweep/Fix] Propagate dynamic type of exception to sweep_device_ge…

MikeGitb:
#### Scope of changes

Fixes error propagation in libsweep from worker to main thread, such that correct dynamic exception is propagated instead of a sliced version


dcyoung:
Thanks for the catch @MikeGitb! 
Tested on Windows and Linux. Seems to work fine. Merging this, and then #103

Issue #104, opened 2017-06-15
RuntimeError: baud rate could not be determined


calebmadrigal:
#### sweep firmware version
1.4

#### libsweep version + affected bindings
Latest from Github (as of June 15, 2017)

#### operating system
OS X 10.10.5

#### Description:

When I try to use the Python SDK (like the simple testing command, `python3 -m sweeppy /dev/cu.usbserial-DO004QG8`), I get this error:

```
cmbpr:sweeppy caleb$ python3 -m sweeppy /dev/cu.usbserial-DO004QG8
Traceback (most recent call last):
  File "/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/runpy.py", line 170, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/caleb/Repos/sweep-sdk/sweeppy/sweeppy/__main__.py", line 26, in <module>
    main()
  File "/Users/caleb/Repos/sweep-sdk/sweeppy/sweeppy/__main__.py", line 12, in main
    with Sweep(dev) as sweep:
  File "/Users/caleb/Repos/sweep-sdk/sweeppy/sweeppy/__init__.py", line 114, in __enter__
    raise _error_to_exception(error)
RuntimeError: baud rate could not be determined
```

Note that:
* I'm using the correct usb serial path - it's the same one that Sweep Visualizer is using (and it works)
* I installed the C and Python SDK

dcyoung:
I verified there is an issue with libsweep on mac. The travis build wasn't catching this as the dummy does not make use of the actual serial communication. 

The issue is with the `getBaud()` method from libsweep's  `unix/serial.cc`. I just pushed a branch `baud-mac-fix` that should fix it. Try it with:
```bash
git clone https://github.com/scanse/sweep-sdk
git checkout baud-mac-fix
# build as normal ...
```

I'll make a PR for this and get it merged into master, as this is blocking mac support for the time being.

If that still doesn't work, you may need to try using the tty version of the port name with the `sweep-sdk`. You can find it with `ls /dev/tty.*`.  Let me know how it goes.

calebmadrigal:
@dcyoung Your fix works for me, thanks for the quick response!

dcyoung:
[...]

PR #106, opened 2017-06-15
[libsweep/Fix] Mac OS Support for getBaud()

dcyoung:
#### Scope of changes
Removes the conditional checks in the `getBaud()` method from libsweep's `unix/serial.cc` in order to support mac OS. See #104 

Additionally, the baud rate check is reduced to a single check for the only supported baud rate of 115200. This reduces code clutter that wasn't really relevant at this time. We can always add things back in the future if we add support for different baud rates. But there are no plans to do so as of now, so it seems nicer to cut the irrelevant code.

#### Known Limitations
Tested on macOS and linux. Both work fine for me.


daniel-j-h:
@dcyoung want to make a feature release v1.2.0 for the macOS support as described in https://github.com/scanse/sweep-sdk/blob/master/RELEASE.md?

dcyoung:
Done

Issue #109, opened 2017-06-30
Non-blocking reads


MikeGitb:
I think it would be helpful, if there was a possibility to check if new scans are available without having to block if there aren't. 

I can think of 3 ways how this could be accomplished

 - Add a `try_read_scan` that fails if no scans are available (fail should not be reported via exception, but via return code
- Add a `is_scan_ready` function that only checks if a scan is in the queue
- Add a timeout parameter to read_scan / add a new read_scan_timeout function with a user configurabel timeout.

MikeGitb:
If I understand correctly, this project is being shut down so I'm closing this issue.

Issue #108, opened 2017-06-30
Time Stamps


MikeGitb:
Can we add time stamps to the scan data? Either one time stamp per scan or for each individual sample? Especailly when rotating with a low frequency, it can become very difficult to accurately merge the lidar data with data from other sources like e.g. a gyroscope.

Ideally, the data should alredy be timestamped in the sensor itself to avoid jitter introduced by the host operating system., but that would probably open a whole bag of synchronization issues. 
However, at least the worker thread in libsweep could create an utc /  unix timestamp upon reception of a message over the serial interface.

daniel-j-h:
Sounds like a duplicate of https://github.com/scanse/sweep-sdk/issues/19?

MikeGitb:
@daniel-j-h: Ah right - sorry. Apparently I didn't have enough coffee this morning ;).

So let's be more specific: I'm currently implementing the timestamps for my private fork of libsweep, but would prefer to do so in a compatible manner. So the question is what should be the return value of a hypothetical function 
 
    ... sweep_scan_get_time(sweep_scan_s scan, int32_t sample);

Personally I'd prefer if it where a value corresponding to "time since unix epoch without leap seconds" in milli- or microseconds. It is the de-facto time standard for `time_t` and `std::chrono::system_clock` and on most systems easily convertible to UTC. The main drawback I see however is that `system_clock` is not a steady clock, so there might be jumps. Also on some embedded systems "time since unix epoch" might actually be unknown, but I think those systems aren't targeted by the library anyway.

daniel-j-h:
Is a timestamp actually needed or is a monotonically increasing steady integer value good enough? Could the frame of reference be the device's start instead of the unix epoch? What are the use-cases for a timestamp here?

MikeGitb:
> What are the use-cases for a timestamp here?

If the Lidar is mounted on a mobile platform, you might want to fuse the data with that from other sensors (e.g. GPS) in order to get absolute positions. The more precisely you can match the measurements from different sensors, the better the result. A monotonic clock would also work, but in my specific usecase, I'd have to convert it to UTC later anyway.

> Is a timestamp actually needed or is a monotonically increasing steady integer value good enough
Are you talking about a packet counter? In that case the answer is definetively no. If that counter is supposed to represent e.g. milliseconds then yes.

dcyoung:
At the very least I think we should add a timestamp associated with a completed scan. 

[...]

PR #110, opened 2017-07-01
Refactor some libsweep internals

MikeGitb:
This is a collection of refactorings concerning the protocol functions and the `sweep_device_accumulate_scans` function. The reason those are in one PR is that they partially depend on each other.

The PRs has two main goals:
- Pushing more code that is concerned with protocol details into the protocol.hpp and .cpp files
- Refactor `sweep_scan` / `sweep_device_accumulate_scans`  in anticipation of adding time stamps and turning it into a dynamically sized datastructure.

### DIFF
--- a/libsweep/src/protocol.cc
+++ b/libsweep/src/protocol.cc
@@ -109,14 +109,13 @@ response_scan_packet_s read_response_scan(serial::device_s serial) {
   return scan;
 }
 
-response_info_motor_ready_s read_response_info_motor_ready(serial::device_s serial, const uint8_t cmd[2]) {
+response_info_motor_ready_s read_response_info_motor_ready(serial::device_s serial) {
   SWEEP_ASSERT(serial);
-  SWEEP_ASSERT(cmd);
 
   response_info_motor_ready_s info;
   serial::device_read(serial, &info, sizeof(info));
 
-  bool ok = info.cmdByte1 == cmd[0] && info.cmdByte2 == cmd[1];
+  bool ok = info.cmdByte1 == MOTOR_READY[0] && info.cmdByte2 == MOTOR_READY[1];
 
   if (!ok)
     throw error{"invalid motor ready response commands"};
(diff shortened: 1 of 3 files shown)


libsweep/src/sweep.cc
```diff
@@ -19,13 +20,25 @@ struct sweep_error {
...
+  sample samples[SWEEP_MAX_SAMPLES];
   int32_t count;
 };
 
+static sample parse_payload(const sweep::protocol::response_scan_packet_s& msg) {
+  sample ret;
+  ret.angle = msg.get_angle_millideg();
+  ret.distance = msg.distance;
+  ret.signal_strength = msg.signal_strength;
+  return ret;
```
daniel-j-h: Should there be `SWEEP_ASSERT`s here to make sure
- signal strength is 0-255
- angle is 0-360 degree
- distance is positive

MikeGitb: I haven't thought about that, but my gut feeling would be no.

If this is a guarantee we want to give the user, then I'd put the asserts it into the public API access functions. If this is meant as a check, if we got plausible data from the device, I'd not use an assert, but rather a normal check (and then either throw an exception or discard the message). Actually, we could add a `is_valid` function that checks the received message for a correct crc and plausible values.

MikeGitb:
I think I've introduced a potential bug: If the worker thread is stopped before the first packet is parsed successfully, the buffer is indexed with a negative index.

I can't do anything about it till tomorrow.

MikeGitb:
OK, I now removed the commit that tried to create a separate sub-loop for receiving the individual samples from one scan. I think the original structure is both more correct and shorter. 

The remaining difference in the workerthread is now mainly that we buffer samples instead of response messages and the adaption to the changes in the protocol functions (e.g. return values instead of out parameteres, named member functions for `has_error` and `is_sync` ).

libsweep/src/protocol.cc
```diff
@@ -55,91 +55,100 @@ void write_command_with_arguments(serial::device_s serial, const uint8_t cmd[2],
...
+  return param;
 }
 
-void read_response_scan(serial::device_s serial, response_scan_packet_s* scan) {
+response_scan_packet_s read_response_scan(serial::device_s serial) {
   SWEEP_ASSERT(serial);
-  SWEEP_ASSERT(scan);
 
-  serial::device_read(serial, scan, sizeof(response_scan_packet_s));
+  response_scan_packet_s scan;
```
dcyoung: To nitpick... we use the term "scan" frequently to mean a collection of scan_packets. So to avoid confusion I would change this variable name.  Perhaps to `packet`.

dcyoung:
@MikeGitb  Sorry for the delay in merging this. I'll try to get the branch tested on linux and mac tomorrow morning, and merged if everything goes smoothly!

dcyoung:
Gave it a whirl on Ubuntu and Mac... seems to work fine. Merging now.

MikeGitb:
@dcyoung: Sounds good.
Regarding features. Any thoughts on #108 and #109? Are those things interesting for anyone but me?

Issue #72, opened 2017-04-26
Creating sweep object hangs


damyvv:
When we connect the sweep to our usb port it opens on /dev/ttyUSB0. The visualizer program works flawless with the sweep. So we took the example.cc file and succesfully compiled the project. But when we run the example with "./Lidar /dev/ttyUSB0" it prints "Constructing sweep device..." and then just hangs. It doesn't throw any error, but it's not proceeding. Looking at your code I would expect it to output its motor speed, but it doesn't do that. So it doesn't get past following line:
[https://github.com/scanse/sweep-sdk/blob/master/libsweep/examples/example.cc#L16](url)

If we disconnect the sweep it throws the "Error: opening serial port failed", so we know "/dev/ttyUSB0" is the right port.

Thanks in advance.

dcyoung:
Hi there damyvv, could you provide a little more info about your setup?

- What operating system are you using?
- What firmware is installed on your sweep device (you can check this using the sweep visualizer)?
- Is `example.cc` the only example causing issues? Does `example.c` work as expected?
- Is the issue with `example.cc` intermittent or does it hang every time you try it?
- Could you clarify how you are calling the example. I'm a little unclear where the `./Lidar` is coming from?

damyvv:
Hi dcyoung,

Thanks for your quick response. I can't access the sweep until monday, but I will answer your questions as soon as possible.

For now I can only provide this information:
- We are using ubuntu version 14.04 (also tried it with ubuntu 16.04 on a different computer, but it had the same result)
- The example.cc hangs every time. Tried it about ten times.
- We compiled the example.cc to an executable and we called it 'Lidar'. So that is just the compiled example.cc file

daniel-j-h:
Can you build both `libsweep` as well as your example with debug information, i.e.

    cmake .. -DCMAKE_BUILD_TYPE=Debug

and then run under `gdb` and give us a backtrace (`bt`) of where exactly it hangs? You can also attach to your running and hanging example program if that makes it easier for you.

If you compiled the [`example.cc`](https://github.com/scanse/sweep-sdk/blob/404a2d65c59e42b2b51c1a4817d8b881c864cee4/libsweep/examples/example.cc#L16) as is it most likely hangs either during serial device construction or sending the initial stop command.

Serial port issues could be related to your user not being allowed to access the serial devices, as in you're not in the `dialout` group:

    sudo usermod -aG dialout $(whoami)

[...]

PR #116, opened 2017-07-17
Fixed sweep object creation hanging

Vrong:
This bug was causing sweep object creation to hang when trying to read the returned value from serial port. ==> Issue #72 

Bitrate of the serial communication was badly set during the creation of the sweep object when using the libsweep library on Linux.

cfsetispeed(...) and cfsetospeed(...) both take a termios bitrate constant and not the bitrate directly. So should the get_baud(115200) call return the B115200 value defined in termios.h.

The fix has been tested successfully and has no limitation.

Thanks.


dcyoung:
Woops! Seems I caused this when I removed the verbose switch statement in commit 51246f731a78ccb927c79f98367ffeedb30d7a72 
Thanks for the catch. 


Any idea why the behavior would affect the RaspberryPi but not a desktop setup? 
And why the issue appeared intermittent? ie: the "hanging" behavior would come and go on the Raspberry Pi. 

Will have to check everything still works on Mac as well. I'll do so now.

dcyoung:
Hmmm I implemented this fix on a Raspberry Pi 3 and witnessed the same issue (still intermittent).  While it may be a good idea to fix `get_baud`, I'm not sure it is the cause of the hanging behavior I'm witnessing. 

Additionally Travis is failing because the check for the definition is missing (removed in 51246f7 for mac compatibility). We'll have to figure out a better solution for portability there.

Vrong:
Ok I will check it again tomorrow as I don't have the lidar now. It was definitely the reason it was not working for me. (Tested on rasperry pi 3 too)
The issue was "intermittent" because another tool or software was sometime setting the 115200 baud bitrate to the serial port.
For example, booting the rasp and using the libsweep API first would hang, but using it after a python serial init would set it to 115200 baud so the libsweep API would work even without succeeding to set bitrate.
You can check the bitrate of the port with sudo stty -F /dev/ttyUSB0 -a
By default I got 9600 bauds, and you can see it's not changing when your soft using the API starts (without the fix).
However the DX write operation would work and I could get the response with a third party serial app (like cutecom or minicom) but not in my software using the API. I guess it was just a lucky serial hardware strike.  
After the fix, I had no single issue as far as I tested it.

Concerning the continuous integration failing, It appears that older versions of termios.h does not define bitrates up to B115200.
`#define  B1152000 0010011`
There is the line but I'm not sure that you could just use the 10011 value and pray for it to be supported. I don't use MAC and did not do more digging on the issue.

That's strange you still have the issue with the fix, maybe that's not the same issue or need to make clean or something ? 0:-)

Well I'll give you more info tomorrow. Thanks a lot for the response.

dcyoung:
Thanks for the info @Vrong  ! I'll look into this more as well.  

Also, what Raspbian release are you using?

dcyoung:
@Vrong I think the intermittent behavior I was seeing can be explained by the sequence of reboots, sweep unplugs and swapping back and forth between libsweep fixes. Now that I'm explicitly checking for it, I can't seem to reproduce the issue with the fix implemented. Checking with `sudo stty -F /dev/ttyUSB0 -a` does seem to validate this fix! 

As for the definition check... I can't remember the exact cause of the mac incompatibility. I'll have to look again, but perhaps a quick and dirty solution could be accomplished rather simply like so:
```c++
#ifdef B115200
    return B115200;
#endif
return 115200;
```

Vrong:
Ok I messed with my repo so I created a new one to keep it clean sorry ==> #117 .

PR #115, opened 2017-07-14
Adds Haskell Bindings

daniel-j-h:
[context](https://www.meetup.com/Berlin-Functional-Programming-Group/events/241451747/) - work in progress

cc @hrantzsch reactormonk @sjsyrek

PR #117, opened 2017-07-18
Fixed badly set bitrate causing sweep object creation to hang on Linux

Vrong:
This bug was causing sweep object creation to hang when trying to read the returned value from serial port. ==> Issue #72

Bitrate of the serial communication was badly set during the creation of the sweep object when using the libsweep library on Linux.

cfsetispeed(...) and cfsetospeed(...) both take a termios bitrate constant and not the bitrate directly. So should the get_baud(115200) call return the B115200 value defined in termios.h.

The fix seems to have a limitation when compiling on some MAC OS X computers. Termios constants like B115200 doesn't seem to be always defined. That's why the code uses a #ifdef as a workaround for MAC.

Thanks.

### DIFF
--- a/libsweep/src/unix/serial.cc
+++ b/libsweep/src/unix/serial.cc
@@ -27,14 +27,10 @@ static speed_t get_baud(int32_t bitrate) {
     return -1;
   }
 
-  // translate human readable bitrate to termios bitrate
-  if (bitrate == 115200)
-  {
-    #ifdef B115200
-    bitrate = B115200;
-    #endif
-  }
-
+// translate human readable bitrate to termios bitrate
+#ifdef B115200
+  return B115200;
+#endif
   return bitrate;
 }
 
@@ -157,7 +153,7 @@ void device_read(device_s serial, void* to, int32_t len) {
         } else {
           throw error{"reading from serial device failed"};
         }
-      } else if(ret == 0){
+      } else if (ret == 0) {
         throw error{"encountered EOF on serial device"};
       } else {
         bytes_read += ret;
--- a/libsweep/src/unix/serial.cc
+++ b/libsweep/src/unix/serial.cc
@@ -26,6 +26,15 @@ static speed_t get_baud(int32_t bitrate) {
     throw error{"Only baud rate 115200 is supported at this time."};
     return -1;
   }
+
+  // translate human readable bitrate to termios bitrate
+  if (bitrate == 115200)
+  {
+    #ifdef B115200
+    bitrate = B115200;
+    #endif
+  }
+
   return bitrate;
 }
 


libsweep/src/unix/serial.cc
```diff
@@ -26,6 +26,15 @@ static speed_t get_baud(int32_t bitrate) {
     throw error{"Only baud rate 115200 is supported at this time."};
     return -1;
   }
+
+  // translate human readable bitrate to termios bitrate
+  if (bitrate == 115200)
```
dcyoung: This conditional branch is redundant given the `if (bitrate != 115200)` above, which results in an error. Maybe just 
```c++
// if defined, translate human readable bitrate to termios bitrate
#ifdef B115200
    return B115200
#endif
return bitrate;
```

libsweep/src/unix/serial.cc
```diff
@@ -26,6 +26,15 @@ static speed_t get_baud(int32_t bitrate) {
...
+  {
+    #ifdef B115200
```
dcyoung: Could you clang format as well. Format file included in the main repo. If not, let me know and I'll clean it up before merging.

dcyoung:
Thanks for this second PR. Even if a more robust and elegant portability fix can't be resolved just yet, the library is currently broken.... so I'd like to get this merged ASAP!!

libsweep/src/unix/serial.cc
Vrong: Well ok, I was supposing that more bitrates would be supported in the future.

libsweep/src/unix/serial.cc
Vrong: I'm not use to it but I can figure out how it works

Vrong:
Done

dcyoung:
I'll do one more final check on mac and get this merged! Thanks so much.

daniel-j-h:
Urgh thanks for the deep dive here and fixing the bug.

I think we hit this last week when playing with Haskell bindings in https://github.com/scanse/sweep-sdk/pull/115 but didn't look further (since struggling with Haskell was already enough for the evening 😅).

libsweep/src/unix/serial.cc
```diff
@@ -148,7 +153,7 @@ void device_read(device_s serial, void* to, int32_t len) {
         } else {
           throw error{"reading from serial device failed"};
         }
-      } else if(ret == 0){
+      } else if (ret == 0) {
```
daniel-j-h: I think this should check for [feof](http://en.cppreference.com/w/cpp/io/c/feof) (?)

libsweep/src/unix/serial.cc
```diff
@@ -26,6 +26,11 @@ static speed_t get_baud(int32_t bitrate) {
     throw error{"Only baud rate 115200 is supported at this time."};
     return -1;
   }
+
+// translate human readable bitrate to termios bitrate
+#ifdef B115200
+  return B115200;
+#endif
```
daniel-j-h: This is a good quick-fix but we should refactor the function (or how we allow for bitrates here).

We take a bitrate argument but in case we hit the branch here the argument is never used.

Vrong:
The first fix I made was actually including a condition testing the bitrate before setting a constant, but this one was not merged.

PR #118, opened 2017-07-28
[Sweeppy Docs] Clarify sample units and format in sweeppy README. 

dcyoung:
We've been receiving quite  a few support messages from people confused by the output data format of the sweeppy example. Specifically, the milli-degree angle value is throwing people off. 

This is a quick patch the sweeppy README  that clarifies the format and units of the returned samples. 

Additionally, this PR updates the libsweep and sweeppy READMEs to include links to 2 articles on using the sweep (theory of operation and best practices).

We should probably also clarify the format and units in the libsweep and sweeppy examples.  Printing a simple header before printing the sample of gathered scans would go a long way.

### DIFF
--- a/libsweep/README.md
+++ b/libsweep/README.md
@@ -87,6 +87,7 @@ Table of Contents:
 - [Error Handling](#error-handling)
 - [Device Interaction](#device-interaction)
 - [Full 360 Degree Scan](#full-360-degree-scan)
+- [Additional Information](#additional-information)
 
 #### Firmware Compatibility
 | libsweep | sweep firmware |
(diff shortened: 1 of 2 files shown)


daniel-j-h:
Hm we probably should have converted to floating point types in all wrappers except the low-level libsweep C API. We still could but that would require a breaking API change:

https://github.com/scanse/sweep-sdk/blob/7dca35cbfa56ac667266d316ad12771cfdcde667/sweeppy/sweeppy/__init__.py#L211-L214

👍 on docs!

dcyoung:
I agree that would be preferable. Lets leave that for a later PR (hopefully soon). And just update the docs until then.

MikeGitb:
If you do  change to floats, I'd personally prefer rad over degree

PR #119, opened 2017-08-07
character size must be set AFTER clearing character size bits

vitorboschi:
Fix another issue in Linux serial port configuration, which causes the issue #72

### DIFF
--- a/libsweep/src/unix/serial.cc
+++ b/libsweep/src/unix/serial.cc
@@ -95,8 +95,8 @@ device_s device_construct(const char* port, int32_t bitrate) {
   // IEXTEN
 
   // Control Flags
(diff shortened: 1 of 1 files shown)


daniel-j-h:
Urgh I fail to see why it worked before then? @dcyoung any ideas here?

dcyoung:
No clue why it worked before this. But judging by the serial guide for posix systems [here](https://www.cmrr.umn.edu/~strupp/serial.html), it looks like this PR is the correct order. If it isn't introducing any unexpected behavior on linux, then I'd advocate we merge.

daniel-j-h:
Just tested this branch with my device on Linux - worked fine.

daniel-j-h:
I just did a new release: https://github.com/scanse/sweep-sdk/releases

Issue #122, opened 2017-10-13
Laggy values from Sweep with piCamera integration in Raspberry pi 3 b


Biswajit1995:
Hey There,

I am having a issue while integrating Sweep code with one of my code that I was using previously.

Actually as I am working with Raspberry pi (Model 3B), I'm more comfortable with Python, and I am using python3 for the development.

My previous code was simple code to take pictures frame by frame from the RaspiCam and use it after some image processing.(Sorry, I can't post the code over here) but it is fairly simple like the below code snippet: 

```
cap = cv2.VideoCapture(camera_index)
cap.set(3, any_resolution)
cap.set(4, any_resolution)
cap.set(5, frame_rate)
```

then under a continuous loop I'm using `cap.read()` to read out frames and passing them through my another set of code to process and use it.

And It was working perfectly fine unless I started to use it with Sweep Lidar code, from the SDK sweeppy I'm using their startup code [(Here)](https://github.com/scanse/sweep-sdk/blob/master/sweeppy/README.md), and since in the Interface scan.get_scans() is returning an Iterable of Sample class which in turn is a class, I'm accessing it through two for loops as mentioned below:

```
from sweeppy import Sweep
with Sweep('port') as sweep:
    sweep.start_scanning()

    for scan in sweep.get_scans():
        for i in scan:
            for j in i:
                print(j.angle/1000, j.distance)
```
and it was perfectly fine without any lag.
The main problem arises when I integrate this piece of code with `cap.read()` something similar to the below one:

```
from sweeppy import Sweep
with Sweep('port') as sweep:
    sweep.start_scanning()

    for scan in sweep.get_scans():
        for i in scan:
            for j in i:
                 flag, frame = cap.read()
                 print(flag)     # To check the camera read is successful or not
                 print(j.angle/1000, j.distance)
```

here, in the output , its only prints True, or false(In case of camera failure), but not printing the j.angle and j.distance values as it was printing before, there is huge lag, means after 1 min 5 values will get printed then again it will print True or False and it goes on like that.

output example:

```
True
True
True
True
True
True
True
True
True
True
True
True
True
(240, 56)
True
(98, 578)
True
(340, 568)
True
(45, 678)
True
True
True
True
True
True
True
True

```
I don't know why it is behaving like that, any help with this would be great.

And one more thing I want to ask is that, is it necessary to use sweeppy SDK with
`with Sweep('dev') as sweep:` part, is there any alternative to this.

Regards
Biswajit

[...]

PR #125, opened 2017-11-07
Adds producer / consumer example for common SweepPy use-case

daniel-j-h:
See https://github.com/scanse/sweep-sdk/issues/122#issuecomment-340136484.

Hope that helps.


daniel-j-h:
For the record: not tested on a real device because I don't have one here atm.

PR #134, opened 2019-08-05
Fix build with GCC

pkubaj:
GCC is used on FreeBSD on e.g. powerpc64.


daniel-j-h:
Looks good; macOS builds are failing Travis because the Travis build environment changed.

daniel-j-h:
Thank you! :bow:

